PR #31, opened 2020-01-15
Move faulty ignore annotations

Slamdunk:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -27,8 +27,8 @@ use stdClass;
  * @method HeadLink prependAlternate($href, $type, $title, $extras = [])
  * @method HeadLink setAlternate($href, $type, $title, $extras = [])
  */
-// @codingStandardsIgnoreEnd
 class HeadLink extends Placeholder\Container\AbstractStandalone
+// @codingStandardsIgnoreEnd
 {
     /**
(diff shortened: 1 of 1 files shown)


src/Helper/HeadLink.php
```diff
@@ -27,7 +26,6 @@
  * @method HeadLink prependAlternate($href, $type, $title, $extras = [])
  * @method HeadLink setAlternate($href, $type, $title, $extras = [])
  */
-// @codingStandardsIgnoreEnd
```
Slamdunk: This annotation is both useless and faulty:

1. Useless because it tells to ignore a docblock
1. Faulty because, placed between the docblock and the class declaration, inhibit PHP to recognize the docblock as a class phpdoc

michalbundyra: How we can then fix these:

image

Slamdunk: I'm so sorry, I misread it as a code coverage annotation.

Now the coding standard annotation is only placed after the class declaration.

weierophinney:
Thanks, @Slamdunk!

PR #39, opened 2020-05-01
Add note about using RendererInterface::render()

aleksip:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Adding a note about the possibility to use `RendererInterface::render()` instead.

### DIFF
--- a/docs/book/helpers/partial.md
+++ b/docs/book/helpers/partial.md
@@ -4,11 +4,6 @@ The `Partial` view helper is used to render a specified template within its own
 variable scope. The primary use is for reusable template fragments with which
 you do not need to worry about variable name clashes.
 
-> Note that it is also possible to render a template within its own variable
-> scope by calling `RendererInterface::render()` with a `$values` argument.
-> See the documentation for [`PhpRenderer`](../php-renderer.md#render) for
-> more information.
-
 A sibling to the `Partial`, the `PartialLoop` view helper allows you to pass
 iterable data, and render a partial for each item.
 
@@ -79,6 +74,14 @@ Which would then render:
 > `Laminas\Db\ResultSet\ResultSet`s to `partialLoop()`, as you then have full
 > access to your row objects within the view scripts, allowing you to call
 > methods on them (such as retrieving values from parent or dependent rows).
+
+> ### An alternative to using Partial
+>
+> Note that it is also possible to render a template within its own variable
+> scope by directly calling `RendererInterface::render()` with a `$values`
+> argument. You might prefer to do this if your model is an array, or an
+> object that implements `ArrayAccess`. See the documentation for
+> [`PhpRenderer`](../php-renderer.md#render) for more information.
 
 ## Using PartialLoop to Render Iterable Models
 
--- a/docs/book/helpers/partial.md
+++ b/docs/book/helpers/partial.md
@@ -4,6 +4,11 @@ The `Partial` view helper is used to render a specified template within its own
 variable scope. The primary use is for reusable template fragments with which
 you do not need to worry about variable name clashes.
 
+> Note that it is also possible to render a template within its own variable
+> scope by calling `RendererInterface::render()` with a `$values` argument.
+> See the documentation for [`PhpRenderer`](../php-renderer.md#render) for
+> more information.
+
 A sibling to the `Partial`, the `PartialLoop` view helper allows you to pass
 iterable data, and render a partial for each item.
 


docs/book/helpers/partial.md
```diff
@@ -4,6 +4,11 @@ The `Partial` view helper is used to render a specified template within its own
 variable scope. The primary use is for reusable template fragments with which
 you do not need to worry about variable name clashes.
 
+> Note that it is also possible to render a template within its own variable
+> scope by calling `RendererInterface::render()` with a `$values` argument.
+> See the documentation for [`PhpRenderer`](../php-renderer.md#render) for
+> more information.
+
```
froschdesign, line 11: This note could be an interesting information for the user but it should be moved to the end and not added in the introduction of this helper. And it should also be highlighted that there is a difference in the behaviour of the method `render` and the partial helper.
Please add also a headline to the blockquote; third order (`###`).

froschdesign (changes requested):
Thank you for this contribution!

Only small changes are necessary so that the information can be included in the documentation.

docs/book/helpers/partial.md
```diff
@@ -80,6 +75,14 @@ Which would then render:
 > access to your row objects within the view scripts, allowing you to call
 > methods on them (such as retrieving values from parent or dependent rows).
 
+> ### An alternative to using Partial
+>
+> Note that it is also possible to render a template within its own variable
+> scope by directly calling `RendererInterface::render()` with a `$values`
+> argument. You might prefer to do this if your model is an array, or an
+> object that implements `ArrayAccess`. See the documentation for
+> [`PhpRenderer`](../php-renderer.md#render) for more information.
```
froschdesign, line 84: > You might prefer to do this if your model is an array, or an object that implements `ArrayAccess`.

The result is the same with the Partial helper and `render` method for this.

But the Partial helper supports some more:

https://github.com/laminas/laminas-view/blob/2e961a1f698593f8db6a71c193e3312783b46b68/src/Helper/Partial.php#L52-L58

aleksip: I originally put the note in the very beginning as I considered it to be quite important for deciding whether to use `Partial` or not.

It seems that historically, when `render()` did not support a `$values` argument, `Partial` was much more useful, but what is left now is just this additional support for handling data. One might even consider the current introduction ("is used to render a specified template within its own variable scope") to be misleading?

aleksip: I'm referring particularly to this commit, which seems to have changed the nature of `Partial`: https://github.com/laminas/laminas-view/commit/e0825a02fd61af8598237e0a71c08396716ee5e0

froschdesign: I would not put so much emphasis on original or old behavior. The user should receive instructions on how to use the helper. That the `render` method is used is an important hint, but it is also important that the helper offers more. (See "What is a model?")

aleksip: Yes, currently the documentation seems to put too much emphasis on the old variable scope feature that has since been added to `RendererInterface` and is now only proxied by `Partial`.

I would even consider whether the model handling functionality that is left is valuable enough to keep `Partial` instead of deprecating it? The `objectKey` feature for example seems to be equivalent to a simple `render($nameOrModel, [$key => $object])` call?

aleksip: Looking into this more I see that `Partial` and its model are used by other parts of the framework, so I am guessing that there are good reasons for keeping the additional model support. And maybe in any case for keeping `Partial` as the recommended way to always render partial templates, since specific functionality can then be easily added later?

As a developer using the framework for the first time, interested in rendering partial templates in their own variable scope, I have however been confused about:

- the difference between `RendererInterface::render()` and `Partial`, and 
- whether I should still use `Partial` if it only seems to add overhead in my specific use case, where I will not be needing the extended model support. 

I am still unsure about the right answers to these questions, and I would love to find the answers in the documentation. This is what I was attempting to do with the PR. If there are additional changes that are needed I am happy to update the PR accordingly, but for the above reasons I would need some more guidance.

froschdesign: I dislike the entire page and the missing code samples for the different model types.

I will rewrite the page:

* explain the internal functioning (`render` method)
* add code samples for all types of models
* [extract the PartialLoop helper to a separate page](https://github.com/laminas/laminas-view/issues/40)
* add relevant links

aleksip: Fantastic, thanks!

froschdesign: No, thank you for bringing up this topic.

---

Btw. please have a look at the projects, there will you find "[Documentation: cookbook recipes](https://github.com/orgs/laminas/projects/8)".
More ideas are needed!

aleksip: Thanks, will have a look!

gsteel:
Pretty confident this patch is abandoned. Please re-open if necessary.

PR #42, opened 2020-05-20
Remove xdebug message on exception

cb0:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

If a non existing form fieldset/element should be rendered, and xdebug is currently enabled, the exception will contain a public property called `xdebug_message` containing html formatted message.
 
That message will be huge (500+MB in my case) depending on the complexity of the form (see screenshot).

https://imgur.com/a/p90IxRR

For most browser this will result in 
- waiting a long time 
- and finally get a blank page as the amount of data is too big

I you don't look at the network manager you'll have no clue what is going on.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
         "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-feed": "^2.7",
         "laminas/laminas-filter": "^2.6.1",
+        "laminas/laminas-form": "^2.14.5",
         "laminas/laminas-http": "^2.5.4",
         "laminas/laminas-i18n": "^2.6",
         "laminas/laminas-log": "^2.7",
--- a/src/Renderer/PhpRenderer.php
+++ b/src/Renderer/PhpRenderer.php
@@ -10,6 +10,7 @@ namespace Laminas\View\Renderer;
 
 use ArrayAccess;
 use Laminas\Filter\FilterChain;
+use Laminas\Form\Exception\InvalidElementException;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\View\Exception;
 use Laminas\View\Helper\AbstractHelper;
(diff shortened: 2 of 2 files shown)


froschdesign:
@cb0 
> If a non existing form fieldset/element should be rendered, and xdebug is currently enabled, the exception will contain a public property called `xdebug_message` containing html formatted message.

Unfortunately I can not reproduce the problem.

In a view script:

```php
/**
 * @var Laminas\Form\Form $form
 */
$form->get('foobar');
```

The result without waiting or huge load:

```
Laminas\Form\Exception\InvalidElementException
No element by the name of [foobar] found in form
```

weierophinney:
I'm unable to reproduce, either.

How I tried:

- I created a simple form in my `IndexController` class:

  ```php
  $search = new Element('search');
  $search->setLabel('search');
  $search->setAttribute('type', 'text');
  $form = new Form('search-form');
  $form->add($search);

  return new ViewModel(['form' => $form]);
  ```

- I updated my `index.phtml` to add the following line:

  ```php
  <?= $form->get('non-existent') ?>
  ```

- I installed xdebug, enabled it, and then configured the following settings:

  ```dosini
  xdebug.default_enable = 1
  xdebug.trace_format = 2
  ```

  In fact, I tried with and without each of those settings, just to verify.

- I fired up the built-in web server: `php -S 0:8080 -t public public/index.php`, and then visited localhost:8080.

I found in each scenario that:

- An exception was thrown, and its stack trace displayed in-page (as I was in development mode).
- The page size was the same with and without xdebug enabled, or with any of the above configuration values set.

My environment:

- PHP 7.4.8
- Xdebug 2.9.6

@cb0 Can you provide us with more information on how you were able to reproduce this? What PHP version and Xdebug version were you using? What Xdebug settings had you configured?

michalbundyra:
Closed due to inactivity.

If the issue is still present, please open another PR and provide more information. Thanks.

PR #53, opened 2020-08-11
Omitting Script attribute if none or default

### LINKED ISSUE #51: type in script tags for classic scripts (js) should be ommited
Referring to:
https://html.spec.whatwg.org/multipage/scripting.html#the-script-element

The type attribute text/javascript is unnecessary and should be ommited.

But it seems, there is no option to get rid of it:

https://github.com/laminas/laminas-view/blob/313e1c7b20094ebe79e12db420f6b793aca93206/src/Helper/HeadScript.php#L174

DennisDobslaf:
Ommiting the script attribute if its either empty or default (text/javascript) and the doctype is set to HTML5.
Referring to: https://html.spec.whatwg.org/multipage/scripting.html#the-script-element

Signed-off-by: Dennis Dobslaf <dennis.dobslaf@gmail.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/src/Helper/HeadScript.php
+++ b/src/Helper/HeadScript.php
@@ -35,6 +35,11 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
     const SCRIPT = 'SCRIPT';
 
     /**
+     * @internal
+     */
+    const DEFAULT_SCRIPT_TYPE = 'text/javascript';
+
+    /**
      * Are arbitrary attributes allowed?
      *
      * @var bool
@@ -130,7 +135,7 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
         $spec = null,
         $placement = 'APPEND',
         array $attrs = [],
-        $type = 'text/javascript'
+        $type = self::DEFAULT_SCRIPT_TYPE
     ) {
         if ((null !== $spec) && is_string($spec)) {
             $action    = ucfirst(strtolower($mode));
@@ -171,7 +176,7 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
 
             $action  = $matches['action'];
             $mode    = strtolower($matches['mode']);
-            $type    = 'text/javascript';
+            $type    = self::DEFAULT_SCRIPT_TYPE;
             $attrs   = [];
 
             if ('offsetSet' == $action) {
@@ -267,7 +272,7 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
      */
     public function captureStart(
         $captureType = Placeholder\Container\AbstractContainer::APPEND,
-        $type = 'text/javascript',
+        $type = self::DEFAULT_SCRIPT_TYPE,
         $attrs = []
     ) {
         if ($this->captureLock) {
@@ -401,7 +406,10 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
         $addScriptEscape = ! (isset($item->attributes['noescape'])
             && filter_var($item->attributes['noescape'], FILTER_VALIDATE_BOOLEAN));
 
-        if (empty($item->type) && $this->view && $this->view->plugin('doctype')->isHtml5()) {
+        if ((empty($item->type) || strtolower($item->type) === self::DEFAULT_SCRIPT_TYPE)
+            && $this->view
+            && $this->view->plugin('doctype')->isHtml5()
+        ) {
             $html = '<script ' . $attrString . '>';
         } else {
             $type = ($this->autoEscape) ? $this->escapeAttribute($item->type) : $item->type;
--- a/test/Helper/HeadScriptTest.php
+++ b/test/Helper/HeadScriptTest.php
@@ -534,4 +534,18 @@ document.write(bar.strlen());');
         $test = $this->helper->__invoke()->toString();
         $this->assertContains('async="', $test);
     }
+
+    /**
+     * @group 23
+     */
+    public function testOmitsTypeAttributeIfNoneGivenAndHtml5Doctype()
+    {
+        $view = new \Laminas\View\Renderer\PhpRenderer();
+        $view->plugin('doctype')->setDoctype(\Laminas\View\Helper\Doctype::HTML5);
+        $this->helper->setView($view);
+
+        $this->helper->__invoke()->appendScript('// some script' . PHP_EOL);
+        $test = $this->helper->__invoke()->toString();
+        $this->assertNotRegExp('#type="text/javascript"#i', $test);
+    }
 }
--- a/src/Helper/InlineScript.php
+++ b/src/Helper/InlineScript.php
@@ -32,7 +32,7 @@ class InlineScript extends HeadScript
         $spec = null,
         $placement = 'APPEND',
         array $attrs = [],
-        $type = 'text/javascript'
+        $type = self::DEFAULT_SCRIPT_TYPE
     ) {
         return parent::__invoke($mode, $spec, $placement, $attrs, $type);
     }


src/Helper/HeadScript.php
```diff
@@ -33,6 +33,7 @@ class HeadScript extends Placeholder\Container\AbstractStandalone
      */
     const FILE   = 'FILE';
     const SCRIPT = 'SCRIPT';
+    const DEFAULT_SCRIPT_TYPE = 'text/javascript';
```
weierophinney, line 36: Let's declare `private` visibility on this one, as it's internal; that way we're not also changing the external API for a bugfix.

test/Helper/HeadScriptTest.php
```diff
@@ -534,4 +534,18 @@ public function testSupportsAsyncAttribute()
...
+     */
+    public function testOmitsTypeAttributeIfNoneGivenAndHtml5Doctype()
+    {
+        $view = new \Laminas\View\Renderer\PhpRenderer();
+        $view->plugin('doctype')->setDoctype(\Laminas\View\Helper\Doctype::HTML5);
+        $this->helper->setView($view);
+
+        $this->helper->__invoke()->appendScript('// some script' . PHP_EOL);
+        $test = $this->helper->__invoke()->toString();
+        $this->assertNotContains('type', $test);
```
weierophinney, line 549: Expand this to `'type="text/javascript"`, to make it absolutely clear what we're trying NOT to emit.

weierophinney (changes requested):
This is great - I just have a couple nit-picks for you to address, and then I can merge.

Thanks!

DennisDobslaf:
@weierophinney The CI failed because of the required PHP Version (private const needs PHP 7.1). If we change this, we cannot require PHP 5.6 anymore. Wouldn't this be a BC Break?

src/Helper/HeadScript.php
DennisDobslaf: private would require PHP 7.1, see comments below. IMO changing PHP Version would be a BC Break

michalbundyra:
@DennisDobslaf we can bump PHP version to 7.2 for the next minor, please see our policy here:
https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2020-03-02-TSC-Minutes.md#lts-policy

We are not considering bumping PHP version in minor update as a BC break.

DennisDobslaf:
> @DennisDobslaf we can bump PHP version to 7.2 for the next minor, please see our policy here:
> https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2020-03-02-TSC-Minutes.md#lts-policy
> 
> We are not considering bumping PHP version in minor update as a BC break.

Well, but this is not a minor update but a fix, isn't it?

So, should it be minor update *with* private const plus PHP 7.2 requirement
or
a fix without private const and leaving the PHP Version as it is?

Should this be a discussion in the slack channel?

samsonasik:
php 5.6, 7, and 7.1 can be removed from `.travis.yml` then

DennisDobslaf:
Any work to do for me now?

froschdesign:
I think there are some misunderstandings. This is a bug fix and the fix should be released with a mini release.
The bump of the PHP version should done separated and within a new minor version. (Btw. the version 7.2 of PHP is [outdated](https://www.php.net/supported-versions.php) and will also be dead in 3 months, so it should be dropped.)

This allows us to make a release now and not to wait for the next minor version.

---

Which means for this pull request:

* remove the visibility modifier from class constant `Laminas\View\Helper\HeadScript\DEFAULT_SCRIPT_TYPE`
* undo all changes in Composer configuration
* undo all changes in Travis configuration
* squashing commits

DennisDobslaf:
> I think there are some misunderstandings. This is a bug fix and the fix should be released with a mini release.
> The bump of the PHP version should done separated and within a new minor version. (Btw. the version 7.2 of PHP is [outdated](https://www.php.net/supported-versions.php) and will also be dead in 3 months, so it should be dropped.)
> 
> This allows us to make a release now and not to wait for the next minor version.
> 
> Which means for this pull request:
> 
> * remove the visibility modifier from class constant `Laminas\View\Helper\HeadScript\DEFAULT_SCRIPT_TYPE`
> * undo all changes in Composer configuration
> * undo all changes in Travis configuration
> * squashing commits

I'll do

froschdesign:
@DennisDobslaf 
Thank you for your understanding and apologies for the trouble!

DennisDobslaf:
@froschdesign I'm with your notes. 
I changed, squashed and pushed now.

src/Helper/HeadScript.php
froschdesign, line 36: This constant should also be used in `Laminas\View\Helper\InlineScript`:

https://github.com/laminas/laminas-view/blob/9512a1939a67cec3deb8e841b651505424f1e168/src/Helper/InlineScript.php#L30-L36

DennisDobslaf: Like `$type = HeadScript::DEFAULT_SCRIPT_TYPE`
or by making a separated "Enum" Class?

froschdesign: The `InlineScript` helper is an extension of the `HeadScript` helper:

https://github.com/laminas/laminas-view/blob/9512a1939a67cec3deb8e841b651505424f1e168/src/Helper/InlineScript.php#L15

Therefore: `$type = self::DEFAULT_SCRIPT_TYPE`

DennisDobslaf: Oh, I've overlooked this fact.

michalbundyra: Probably, as we cannot add visibility on the constant, we should add php annotation `@internal` for now. Then in next minor, when we bump php version we can use private/protected visibility.

DennisDobslaf: added your suggestion @michalbundyra

src/Helper/HeadScript.php
```diff
@@ -401,7 +406,10 @@ public function itemToString($item, $indent, $escapeStart, $escapeEnd)
         $addScriptEscape = ! (isset($item->attributes['noescape'])
             && filter_var($item->attributes['noescape'], FILTER_VALIDATE_BOOLEAN));
 
-        if (empty($item->type) && $this->view && $this->view->plugin('doctype')->isHtml5()) {
+        if ((empty($item->type) || $item->type === self::DEFAULT_SCRIPT_TYPE)
```
michalbundyra, line 409: I think check should be case insensitive, so we should use `strtolower($item->type)` instead. What do you think?

DennisDobslaf:
Seems very legit. I will change it and expand the unit test.

Michał Bundyra <notifications@github.com> schrieb am Di., 18. Aug. 2020,
15:32:

src/Helper/HeadScript.php
DennisDobslaf: Added strtolower and a new assertion to the unit test

test/Helper/HeadScriptTest.php
```diff
@@ -534,4 +534,19 @@ public function testSupportsAsyncAttribute()
...
+    public function testOmitsTypeAttributeIfNoneGivenAndHtml5Doctype()
+    {
+        $view = new \Laminas\View\Renderer\PhpRenderer();
+        $view->plugin('doctype')->setDoctype(\Laminas\View\Helper\Doctype::HTML5);
+        $this->helper->setView($view);
+
+        $this->helper->__invoke()->appendScript('// some script' . PHP_EOL);
+        $test = $this->helper->__invoke()->toString();
+        $this->assertNotContains('type="text/javascript"', $test);
+        $this->assertNotContains('type="TEXT/JAVASCRIPT"', $test);
```
weierophinney, line 550:
```suggestion
        $this->assertNotRegExp('#type="text/javascript#i"', $test);
```

A case-insensitive regexp is better here, as it ensures no case variation of that character sequence will be present.

weierophinney (changes requested):
One final change, and I think this is ready!

test/Helper/HeadScriptTest.php
DennisDobslaf: Thanks for your advises. I updated the test (+set your " to the correct position ;-) )

samsonasik (approved):

samsonasik:
@weierophinney I think it is ready now

michalbundyra (approved):

DennisDobslaf:
Any news on this one? Do I have to do some work?

froschdesign (approved):

samsonasik:
ping @weierophinney

froschdesign:
@samsonasik 
You can merge it yourself! No need to ping the project leader here. 😉

samsonasik:
Merged 🎉  @DennisDobslaf thank you for the PR ;)
@froschdesign thank you for the review ;)

PR #58, opened 2020-10-07
PHP 8.0 support

hummer2k:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

This PR adds PHP 8.0 support. 
Phpunit has been updated to version 9.3
Tests have been updated.

See #57

PR #59, opened 2020-10-13
Doctype ViewHelper Factory

### LINKED ISSUE #50: Doctype doesn't have a factory in Mezzio
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.11.4

#### Summary

The `Doctype` view helper does not have a factory in Mezzio, so configuration is not respected.

#### Current behavior

When invoking the Doctype view helper in a Mezzio app with `<?= $this->doctype() ?>` in the view, there is no special factory for it in Mezzio, so the `InvokableFactory` kicks in.
Therefore the configuration as in `laminas-mvc` does not work:
```
'view_manager' => [
    'doctype' => 'HTML5',
],
```

That starts to become a problem when using the `HeadMeta` helper with a `property` key type, because the default Doctype is `Doctype::HTML4_LOOSE` and `property` [is not allowed in that case](https://github.com/laminas/laminas-view/blob/master/src/Helper/HeadMeta.php#L352).

In `laminas-mvc` it is circumvented by injecting a [custom factory](https://github.com/laminas/laminas-mvc/blob/master/src/Service/ViewHelperManagerFactory.php#L71). But this factory is in the `laminas-mvc` package and therefore not available in Mezzio.

#### How to reproduce

When using the `laminas-view` templating system in Mezzio, add the following in a view file:
```
$this->headMeta()->setProperty('og:title', 'Test');
```

#### Expected behavior

The setting in the config should be respected.

Currently I work around this with a self-written factory:

```
<?php

declare(strict_types=1);

namespace App\View\Factory;

use Laminas\View\Helper\Doctype;
use Psr\Container\ContainerInterface;

final class DoctypeHelperFactory
{
    public function __invoke(ContainerInterface $container): Doctype
    {
        $helper = new Doctype();
        $doctype = $this->config($container)['doctype'];
        $helper->setDoctype($doctype);
        return $helper;
    }

    /**
     * @return array<string, mixed>
     */
    private function config(ContainerInterface $container): array
    {
        $config = $container->get('config');
        return $config['view_manager'];
    }
}
```

... but that should probably handled directly in `laminas-view` to have consistent behavior in MVC and Mezzio.

It seems that this is one of the few remaining tight couplings from pre ZendFramework 2.5 version.

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

This PR adds a factory for the `Doctype` view helper, to be able to configure the used doctype via configuration.
The factory is registered with the default factories in the `HelperPluginManager`.
Also, a test checks if the default doctype is HTML4_LOOSE.

fixes #50

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,6 @@
     "config": {
         "sort-packages": true
     },
-    "extra": {
-    },
     "require": {
         "php": "^7.3 || ~8.0.0",
         "laminas/laminas-eventmanager": "^3.0",
@@ -47,12 +45,15 @@
         "laminas/laminas-permissions-acl": "^2.6",
         "laminas/laminas-router": "^3.0.1",
         "laminas/laminas-serializer": "^2.6.1",
-        "laminas/laminas-servicemanager": "^3.0.3",
+        "laminas/laminas-servicemanager": "^3.3",
         "laminas/laminas-session": "^2.8.1",
         "laminas/laminas-uri": "^2.5",
         "phpspec/prophecy-phpunit": "^2.0",
         "phpspec/prophecy": "^1.12",
         "phpunit/phpunit": "^9.3"
+    },
+    "conflict": {
+        "laminas/laminas-servicemanager": "<3.3"
     },
     "suggest": {
         "laminas/laminas-authentication": "Laminas\\Authentication component",
--- a/docs/book/helpers/doctype.md
+++ b/docs/book/helpers/doctype.md
@@ -45,11 +45,31 @@ And then print it out on top of your layout script:
 <?php echo $this->doctype() ?>
 ```
 
-Within an application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton),
-you can specify the doctype via configuration:
+## Usage in a Mezzio Application
+
+The factory `Laminas\View\Helper\Service\DoctypeFactory` checks the application configuration, making it possible to
+define the doctype through your configuration, e.g. `config/autoload/mezzio.global.php`
+or a `ConfigProvider.php` in a module.
+
+For example, add the following lines to your `config/autoload/mezzio.global.php` file to set the `Doctype` to HTML5:
 
 ```php
-// module/Application/config/module.config.php:
+return [
+    /* ... */
+    'view_helper_config' => [
+        'doctype' => \Laminas\View\Helper\Doctype::HTML5,
+    ],
+];
+```
+
+## Usage in a laminas-mvc Application
+
+If you're running a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, you should specify doctype via the
+[ViewManager](https://docs.laminas.dev/laminas-mvc/services/#viewmanager) service.
+
+Add the following lines to your `config/autoload/global.php` file to set the `Doctype` to HTML5:
+
+```php
 return [
     /* ... */
     'view_manager' => [
@@ -128,8 +148,3 @@ Here is how you check if the doctype is set to `XHTML1_RDFA`:
     <?php endif; ?>
 >
 ```
-
-## Laminas MVC View Manager
-
-If you're running a LaminasMvc application, you should specify doctype via the
-[ViewManager](https://docs.laminas.dev/laminas-mvc/services/#viewmanager) service.
--- a/docs/book/helpers/doctype.md
+++ b/docs/book/helpers/doctype.md
@@ -73,7 +73,7 @@ Add the following lines to your `config/autoload/global.php` file to set the `Do
 return [
     /* ... */
     'view_manager' => [
-        'doctype' => 'html5',
+        'doctype' => \Laminas\View\Helper\Doctype::HTML5,
         /* ... */
     ],
 ];
--- a/test/Helper/Service/DoctypeFactoryTest.php
+++ b/test/Helper/Service/DoctypeFactoryTest.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * @see       https://github.com/laminas/laminas-view for the canonical source repository
+ * @copyright https://github.com/laminas/laminas-view/blob/master/COPYRIGHT.md
+ * @license   https://github.com/laminas/laminas-view/blob/master/LICENSE.md New BSD License
+ */
+
+namespace LaminasTest\View\Helper\Service;
+
+use Interop\Container\ContainerInterface;
+use Laminas\View\Helper\Doctype;
+use Laminas\View\Helper\Service\DoctypeFactory;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+final class DoctypeFactoryTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        Doctype::unsetDoctypeRegistry();
+    }
+
+    public function testServiceIsCreatedOnInvocation(): void
+    {
+        $container = $this->createContainer();
+
+        $factory = new DoctypeFactory();
+        $service = $factory($container, '');
+
+        self::assertInstanceOf(Doctype::class, $service);
+    }
+
+    public function testFactorySetsDoctypeBasedOnConfig(): void
+    {
+        $config    = ['view_helper_config' => ['doctype' => Doctype::XHTML1_STRICT]];
+        $container = $this->createContainer($config);
+
+        $factory = new DoctypeFactory();
+        $service = $factory($container, '');
+
+        self::assertSame(Doctype::XHTML1_STRICT, $service->getDoctype());
+    }
+
+    public function testDefaultDoctypeIsUsedIfConfigIsMissing(): void
+    {
+        $config    = ['view_helper_config' => []];
+        $container = $this->createContainer($config);
+
+        $factory = new DoctypeFactory();
+        $service = $factory($container, '');
+
+        self::assertSame(Doctype::HTML4_LOOSE, $service->getDoctype());
+    }
+
+    /**
+     * @param array<string, mixed> $config
+     * @return ContainerInterface & MockObject
+     */
+    private function createContainer(array $config = [])
+    {
+        $container = $this->createMock(ContainerInterface::class);
+        $container->method('has')->willReturn(true);
+        $container->method('get')->with('config')->willReturn($config);
+        return $container;
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
+.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,11 @@ matrix:
     - php: 7.4
       env:
         - DEPS=latest
-    - php: nightly
+    - php: 8.0
       env:
         - DEPS=lowest
         - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
-    - php: nightly
+    - php: 8.0
       env:
         - DEPS=latest
         - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ## 2.12.0 - TBD
 
+### Added
+
+- [#59](https://github.com/laminas/laminas-view/pull/59) Adds a `Doctype` factory which applies configuration provided via the `view_helper_config` config key in **non Laminas MVC projects**.
+
 - [#58](https://github.com/laminas/laminas-view/pull/58) Adds PHP 8.0 support
 
 ### Changed
--- a/src/Helper/Service/DoctypeFactory.php
+++ b/src/Helper/Service/DoctypeFactory.php
@@ -8,23 +8,12 @@
 
 namespace Laminas\View\Helper\Service;
 
-use Interop\Container\ContainerInterface;
-use Laminas\ServiceManager\Factory\FactoryInterface;
-use Laminas\View\Exception;
 use Laminas\View\Helper\Doctype;
+use Psr\Container\ContainerInterface;
 
-class DoctypeFactory implements FactoryInterface
+final class DoctypeFactory
 {
-    /**
-     * {@inheritDoc}
-     *
-     * @param ContainerInterface $container
-     * @param string             $name
-     * @param null|array         $options
-     * @return Doctype
-     * @throws Exception\RuntimeException
-     */
-    public function __invoke(ContainerInterface $container, $name, array $options = null)
+    public function __invoke(ContainerInterface $container): Doctype
     {
         $helper = new Doctype();
 
--- a/src/Helper/Service/DoctypeFactory.php
+++ b/src/Helper/Service/DoctypeFactory.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * @see       https://github.com/laminas/laminas-view for the canonical source repository
+ * @copyright https://github.com/laminas/laminas-view/blob/master/COPYRIGHT.md
+ * @license   https://github.com/laminas/laminas-view/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Laminas\View\Helper\Service;
+
+use Interop\Container\ContainerInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\View\Exception;
+use Laminas\View\Helper\Doctype;
+
+class DoctypeFactory implements FactoryInterface
+{
+    /**
+     * {@inheritDoc}
+     *
+     * @param ContainerInterface $container
+     * @param string             $name
+     * @param null|array         $options
+     * @return Doctype
+     * @throws Exception\RuntimeException
+     */
+    public function __invoke(ContainerInterface $container, $name, array $options = null)
+    {
+        $helper = new Doctype();
+
+        if (! $container->has('config')) {
+            return $helper;
+        }
+        $config = $container->get('config');
+        if (isset($config['view_helper_config']['doctype'])) {
+            $helper->setDoctype($config['view_helper_config']['doctype']);
+        }
+
+        return $helper;
+    }
+}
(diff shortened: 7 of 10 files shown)


composer.json
```diff
@@ -19,6 +19,9 @@
         "sort-packages": true
     },
     "extra": {
+        "branch-alias": {
+            "dev-doctype-factory": "2.12.x-dev"
+        }
```
rieschl: That workaround fixes the [failure in Travis](https://travis-ci.com/github/laminas/laminas-view/jobs/399039557#L236). But that's specific to the current feature branch. I have no idea how to fix that.

rieschl: solved through  [`COMPOSER_ROOT_VERSION`](https://github.com/laminas/laminas-view/pull/59/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R10) env variable

docs/book/helpers/doctype.md
```diff
@@ -26,39 +26,58 @@ You can also specify a custom doctype as long as it is well-formed.
...
+An application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton)
+comes with its own factory in the `ViewHelperManager` and can be configured like this:
 
 ```php
 // module/Application/config/module.config.php:
 return [
     /* ... */
     'view_manager' => [
-        'doctype' => 'html5',
+        'doctype' => 'HTML5',
```
rieschl: When writing the docs I noticed that the doctype has to be in uppercase, even in `laminas-mvc`.

@weierophinney @froschdesign Is it okay if I change that here or do you want me to change that in a separate PR? Or maybe even change `Doctype` itself to accept lowercase?

froschdesign: @rieschl
> Is it okay if I change that here…

Using a separate commit on this pull request is enough. Btw. good catch! 👍 

> Or maybe even change Doctype itself to accept lowercase?

Good idea! Create an issue report or a pull request – both are welcome.

rieschl: Thanks for the quick reply!


> @rieschl
> > Is it okay if I change that here…
> 
> Using a separate commit on this pull request is enough.

Will do!

> Btw. good catch! 👍 

Thanks. finding and fixing bugs, code reviewing and keeping code quality (tools) up to date are my main tasks in my biggest project, so I kind of have an eye for things like that 🙂


> 
> > Or maybe even change Doctype itself to accept lowercase?
> 
> Good idea! Create an issue report or a pull request – both are welcome.
> 

okay I'll come up with a PR, but probably not today anymore.

rieschl:
Btw. I also tried it directly in a Mezzio project. Works as expected.

docs/book/helpers/doctype.md
rieschl: > 
> Using a separate commit on this pull request is enough.

Done :)

froschdesign:
@rieschl 
I think we should wait for the PHP 8 support and the remove of laminas-servicemanager with version 2.

rieschl:
> 
> 
> @rieschl
> I think we should wait for the PHP 8 support and the remove of laminas-servicemanager with version 2.

Okay, should I do that? PHP8 and dropping servicemanager v2 in one PR?

froschdesign:
@rieschl 
> PHP8 and dropping servicemanager v2 in one PR?

Yes, version 2 of laminas-servicemanager is no longer supported and does not support PHP 8.
Please have a look at #57 and #58.

Thanks in advance! 👍

rieschl:
> 
> 
> @rieschl
> 
> > PHP8 and dropping servicemanager v2 in one PR?
> 
> Yes, version 2 of laminas-servicemanager is no longer supported and does not support PHP 8.
> Please have a look at #57 and #58.
> 
> Thanks in advance! 👍

Ah, I didn't notice that there is already a PR for PHP8. Thanks!

docs/book/helpers/doctype.md
```diff
@@ -26,39 +26,58 @@ You can also specify a custom doctype as long as it is well-formed.
...
 [Placeholder helper](placeholder.md).
 
-## Basic Usage
+## Configuration and Basic Usage
 
 You may specify the doctype at any time. However, helpers that depend on the
-doctype for their output will recognize it only after you have set it, so the
-easiest approach is to specify it in your bootstrap:
+doctype for their output will recognize it only after you have set it.
+There are multiple ways to set the doctype:
```
froschdesign: This section should be reverted because "Basic Usage" is the first correct entry. The previous version was okay.

docs/book/helpers/doctype.md
```diff
@@ -26,39 +26,58 @@ You can also specify a custom doctype as long as it is well-formed.
...
+    ],
+];
 ```
 
-Within an application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton),
-you can specify the doctype via configuration:
+### `laminas-mvc` configuration
+
+An application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton)
+comes with its own factory in the `ViewHelperManager` and can be configured like this:
```
froschdesign: The headline should be "Usage in a laminas-mvc Application" and the second level should be used.

froschdesign: The link to the "ViewManager" from the end of the page can be added here. [The last headline and paragraph](https://docs.laminas.dev/laminas-view/helpers/doctype/#laminas-mvc-view-manager) of the current page can be removed then.

docs/book/helpers/doctype.md
```diff
@@ -26,39 +26,58 @@ You can also specify a custom doctype as long as it is well-formed.
...
+`\Laminas\View\Helper\Service\DoctypeFactory` checks the application configuration, making it possible to
+define the doctype through your configuration (e.g. a `ConfigProvider`):
 
 ```php
-<?php echo $this->doctype() ?>
+return [
+    'view_helper_config' => [
+        'doctype' => 'HTML5',
+    ],
+];
```
froschdesign: The headline should be "Usage in a Mezzio Application" and the second should level be used. The entire section should describe the usage in a Mezzio application.

docs/book/helpers/doctype.md
rieschl: Done! I borrowed the headline from the `asset` documentation 🙂 

Maybe you should come up with a Documentation guide, like the [Maintainers guide](https://github.com/laminas/technical-steering-committee/blob/main/MAINTAINERS.md) 😉

docs/book/helpers/doctype.md
rieschl: done!

docs/book/helpers/doctype.md
rieschl: done

rieschl:
@froschdesign , thank you for your feedback! When the [PHP8 PR](https://github.com/laminas/laminas-view/pull/58) is merged, I'll update this PR and remove ServiceManager v2 stuff.

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,30 @@ And then print it out on top of your layout script:
...
+
+```php
+return [
+    'view_helper_config' => [
+        'doctype' => 'HTML5',
+    ],
+];
+```
+
+## Laminas MVC View Manager
```
froschdesign:
```suggestion
## Usage in a laminas-mvc Application
```
We already use this wording in other places.

rieschl: argh damn. sorry, I already had it and then pasted the heading from the last paragraph over it 🤦‍♂️

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,30 @@ And then print it out on top of your layout script:
...
+return [
+    'view_helper_config' => [
+        'doctype' => 'HTML5',
+    ],
+];
+```
+
+## Usage in a laminas-mvc Application
+
+If you're running a [LaminasMvc](https://github.com/laminas/laminas-mvc-skeleton) application, you should specify doctype via the
```
froschdesign:
```suggestion
If you're running a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, you should specify doctype via the
```

boesing:
@rieschl I've rebased against latest 2.12.x (I've merged #58). Please apply your changes regarding SM v2 so we can release 2.12.0 after this 👍

rieschl:
> 
> 
> @rieschl I've rebased against latest 2.12.x (I've merged #58). Please apply your changes regarding SM v2 so we can release 2.12.0 after this 👍

Thanks, @boesing ! I removed the v2 compatibility
I also added `.phpunit.result.cache` to the gitignore. It seems to have been forgotten.

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,30 @@ And then print it out on top of your layout script:
...
+If you're running a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, you should specify doctype via the
+[ViewManager](https://docs.laminas.dev/laminas-mvc/services/#viewmanager) service.
 
 ```php
 // module/Application/config/module.config.php:
 return [
     /* ... */
     'view_manager' => [
-        'doctype' => 'html5',
+        'doctype' => 'HTML5',
```
froschdesign:
```suggestion
        'doctype' => Laminas\View\Helper\Doctype::HTML5,
```

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,30 @@ And then print it out on top of your layout script:
 <?php echo $this->doctype() ?>
 ```
 
-Within an application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton),
-you can specify the doctype via configuration:
+## Usage in a Mezzio Application
+
+`\Laminas\View\Helper\Service\DoctypeFactory` checks the application configuration, making it possible to
+define the doctype through your configuration (e.g. a `ConfigProvider`):
```
froschdesign:
```suggestion
The factory `Laminas\View\Helper\Service\DoctypeFactory` checks the application configuration, making it possible to define the doctype through your configuration, e.g. `config/autoload/mezzio.global.php`:
```

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,30 @@ And then print it out on top of your layout script:
...
+
+```php
+return [
+    'view_helper_config' => [
+        'doctype' => 'HTML5',
```
froschdesign:
```suggestion
        'doctype' => Laminas\View\Helper\Doctype::HTML5,
```

froschdesign (changes requested):
In a Mezzio application the global configuration can be used and also the class constants of `Laminas\View\Helper\Doctype`.

Thanks in advance!

rieschl:
> In a Mezzio application the global configuration can be used and also the class constants of `Laminas\View\Helper\Doctype`.


Thanks, again 😄 
I updated the docs

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,32 @@ And then print it out on top of your layout script:
...
-Within an application based off the [laminas-mvc-skeleton](https://github.com/laminas/laminas-mvc-skeleton),
-you can specify the doctype via configuration:
+## Usage in a Mezzio Application
+
+The factory `Laminas\View\Helper\Service\DoctypeFactory` checks the application configuration, making it possible to
+define the doctype through your configuration, e.g. `config/autoload/mezzio.global.php`
+or a `ConfigProvider.php` in a module.
+
+```php
+// config/autoload/mezzio.global.php
```
froschdesign: Please add the filename before the code example because there are always two parts for this information: what kind of file is this and what is the name or path.

Examples are:

> Create a file for the module configuration called `module.config.php` under `laminas-mvc-tutorial/module/Album/config/`:

> …in a factory, e.g. `src/Album/Controller/AlbumControllerFactory.php`:

> …in a module configuration file, e.g. `module/Album/config/module.config.php`:

In a code comment this would be too long and misplaced.

Thanks in advance!

_(@settermjd and I are still working on documentation writer's guide and we hope to publish it soon.)_

rieschl: done, thanks!

I'm looking forward to a writer's guide! Then hopefully I'll get it right one the first try 🙂

docs/book/helpers/doctype.md
```diff
@@ -45,15 +45,35 @@ And then print it out on top of your layout script:
...
+    ],
+];
+```
+
+## Usage in a laminas-mvc Application
+
+If you're running a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, you should specify doctype via the
+[ViewManager](https://docs.laminas.dev/laminas-mvc/services/#viewmanager) service.
+
+Add the following lines to your `module/Application/config/module.config.php` file to set the `Doctype` to HTML5:
```
froschdesign:
```suggestion
Add the following lines to your `config/autoload/global.php` file to set the `Doctype` to HTML5:
```

This configuration is related to the entire application and not to a module. The same is already done in the previous section for Mezzio. (This is also wrong in the [skeleton application](https://github.com/laminas/laminas-mvc-skeleton/blob/12ff936b69caf63797b9c06a9f3a5d95482be913/module/Application/config/module.config.php#L48-L50).)

test/Helper/Service/DoctypeFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+final class DoctypeFactoryTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        Doctype::unsetDoctypeRegistry();
+    }
+
+    public function testServiceIsCreatedOnInvocation(): void
+    {
+        $container = $this->createContainer();
```
froschdesign, line 26: A mock object should be used here which is build by PHPUnit. The interface of the container is relevant here, not the concrete implementation like the service manager.
This saves us work in the future.

Thanks in advance!

froschdesign (changes requested):
Looks good so far! 👍 

I have just two small comments.

test/Helper/Service/DoctypeFactoryTest.php
rieschl: > A mock object should be used here which is build by PHPUnit. The interface of the container is relevant here, not the concrete implementation like the service manager.
> This saves us work in the future.

Really? I almost never use mocks because of the [headache they caused](https://thephp.cc/dates/2020/06/php-fwdays-special-edition/avoiding-mockageddon) me in the past and always use real stuff, if possible. If the ServiceManager or Container changes, the test _should_ break because the Factory won't work in a real application anymore.
But if Mocks are the Laminas guidelines, I'll of course change the code accordingly.

On a side note: Personally, I even think that the whole mocking stuff should be refactored out of PHPUnit, just like Prophecy.

froschdesign: > I almost never use mocks because of the headache they caused…

1. The dependency is the container interface, not the service manager.
2. There is no _overmocking_ in this case.

```php
$container->has('config')
```

https://github.com/laminas/laminas-view/pull/59/files#diff-e6d323bcbed448c76f7d30b10217611dc0ffd9ab40ae27272dd9fe17d87da035R31

```php
$container->get('config')
```
https://github.com/laminas/laminas-view/pull/59/files#diff-e6d323bcbed448c76f7d30b10217611dc0ffd9ab40ae27272dd9fe17d87da035R34

> If the ServiceManager or Container changes, the test should break because the Factory won't work in a real application anymore.

Create a mock object for the interface of the container then it does not matter if the service manager or the container changes.

> Personally, I even think that the whole mocking stuff should be refactored out of PHPUnit, just like Prophecy.

This is the wrong issue tracker! 😉

test/Helper/Service/DoctypeFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $service = $factory($container, '');
+
+        self::assertSame(Doctype::HTML4_LOOSE, $service->getDoctype());
+    }
+
+    /**
+     * @param array<string, mixed> $config
+     * @return ContainerInterface
+     */
+    private function createContainer(array $config = [])
```
boesing, line 60: > I almost never use mocks

That is your personal decision and you can do this in your projects. Thats totally fine if you are doing this.
However, in almost every laminas package, we are using mocks instead of concrete implementations as it solves multiple issues:
1. If a bug rises in `laminas-servicemanager`, this package is not affected directly
2. "Unit tests" are supposed to test a specific unit and not how they interact with their dependencies. Since there is no concrete dependency to the `servicemanager`, there is no need to use the servicemanager as a dependency
3. You are explicitly testing the behavior on how the unit is interacting with the container, without using a mock, you cannot verify if the config was called from the container of if it was assembled within the method you are calling

So, please change this to a mock as it would provide better maintainability of this package. 👍

boesing (changes requested):

docs/book/helpers/doctype.md
rieschl: Thanks! Yes, I was always wondering why global stuff was in the `module` config. Do you want me to submit a PR against laminas-mvc-skeleton to fix that?

test/Helper/Service/DoctypeFactoryTest.php
rieschl: done! I changed the test to use a Mock object

test/Helper/Service/DoctypeFactoryTest.php
rieschl: Thanks for the input! I changed the test to use a Mock. But I still don't really get why a Mock is better than a "real" object, but that's not something to be discussed here 🙂 Maybe sometime in Slack with a (virtual) beer 😉

rieschl:
Okay, I cleaned up the commit history a little bit. Is there anything for me to do?

edit: I added a separate commit at the end which adds the COMPOSER_ROOT_VERSION env to travis. When it's merged, the latest commit can be dumped, according to https://github.com/laminas/laminas-view/pull/58#issuecomment-714539267 . Is that ok, @boesing ?

edit2: do you want me to add an entry in CHANGELOG or is that done when merging the PR?

froschdesign (approved):

boesing:
@rieschl Mostly, maintainers are adding CHANGELOG as its quite hard to determine if a CHANGELOG entry is necessary or not.

e.g. when fixing documentation, there is no need for a dedicated entry within the changelog. there might be other changes (such as inline-phpdoc stuff, e.g.) which does not either.

So its absolutely fine when you omit the CHANGELOG and let maintainers do so 👍🏼 

I'll prepare a release now.

boesing (approved):

boesing:
@froschdesign @Xerkus Should we think about dropping `view_manager` config in `laminas-mvc` in favor of the `view_helper_config` key?
Thinking about some kind of forward compatibility to enable users to prepare for MVC v4.

boesing:
Thanks, @rieschl!

PR #60, opened 2020-12-07
Add Attributes helper

aleksip:
Signed-off-by: Aleksi Peebles <aleksi@iki.fi>

|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

A new helper to print out HTML tag attributes as [discussed in the forum](https://discourse.laminas.dev/t/rfc-helper-and-or-class-to-manipulate-and-print-html-element-attributes/1888/2).

This is a minimal implementation, basically just moving existing methods from `AbstractHtmlElement` to a trait, used by a separate helper for more general use.


src/Helper/AttributesTrait.php
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param array $attribs From this array, each key-value pair is
+     * converted to an attribute name and value.
+     *
+     * @return string The XHTML for the attributes.
+     */
+    protected function htmlAttribs($attribs)
+    {
+        $xhtml          = '';
+        $escaper        = $this->getView()->plugin('escapehtml');
```
froschdesign: Unfortunately that won't work. There is no method `getView` in this trait.

This example will fail:

```php
class Example
{
	use AttributesTrait;

	public function test()
	{
		$this->htmlAttribs([]);
	}
}
```
This means we have a wrong usage of a trait here.

aleksip: So the fact that the trait is meant to be used by view helpers only and is in the `Laminas\View\Helper` namespace does not help (pun intended)? 🙂 

I will try to think of another approach then. Suggestions are welcome!

aleksip: @froschdesign Maybe I could place these in a class extending `AbstractHelper` and extended by `AbstractHtmlElement`?

weierophinney: If we add it as an abstract method to the trait, this should work, however.

aleksip: I was actually just thinking the same and have now done that!

froschdesign:
At the moment this is a very basic helper without any further functions. I missed translation and validation of attributes like in [`Laminas\Form\View\Helper\AbstractHelper`](https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/src/View/Helper/AbstractHelper.php#L33).

aleksip:
@froschdesign This could indeed be developed a bit further. I have some ideas I am working on, so marked this as draft for now.

aleksip:
@froschdesign @weierophinney I have now added an `AttributeStore` class which I have been dogfooding successfully. This is still basic functionality for use cases where that is all that is needed. Validation and translation could be added, of course. Just checking at this point whether `AttributeStore` and the implementation are acceptable?

froschdesign:
@aleksip 
Thanks for your suggestion!

Some short comments:

* the duplication in the naming: `AttributeStore::setAttributes()` (see also in the [forum](https://discourse.laminas.dev/t/rfc-helper-and-or-class-to-manipulate-and-print-html-element-attributes/1888/2?u=froschdesign))
* Can and should the new class `AttributeStore` be used for other things? Or are there other reasons why the functionality is extracted?

aleksip:
@froschdesign 

- The methods for single attributes could indeed be shortened to just `set()`, `add()` etc.! Should the methods for all attributes then be `setAll()` etc.?
- Unless I have misunderstood something there is only one instance of a specific helper per request? So a separate store class is useful when a template needs to process attributes for several tags. I thought this might be a more straightforward approach than managing several sets of attributes in the helper itself. `AttributeStore` could basically be a rather independent class, but if we want to make use of the helper plugin system, there needs to be at least an optional coupling.

froschdesign:
> So a separate store class is useful when a template needs to process attributes for several tags.

A simple `array` or PHP's `ArrayObject` can be used here.

I still think that a real added value would be the functions from the [abstract view helper of laminas-form](https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/src/View/Helper/AbstractHelper.php#L33). Enhanced with the support of application wide configuration for custom attributes.

aleksip:
My original implementation was array-based, but an `$attributes->add('class', 'active')` seems much cleaner to me than `$attributes = $this->attributes->add($attributes, 'class', 'active')`. Or are you thinking of some other way?

I can assure that just this simple functionality brings **huge** value for the style of templates that are common with Drupal, at least! But the functions from laminas-form would be a great addition too.

froschdesign:
> …an `$attributes->add('class', 'active')` seems much cleaner to me than `$attributes = $this->attributes->add($attributes, 'class', 'active')`. Or are you thinking of some other way?

No it is but the storing for multiple usage of the same attributes can be done via `array` or `ArrayObject`.

aleksip:
Do you mean something like `$this->attributes->add('someId', 'class', 'active')` and then `$this->attributes->toString('someId')`?

I considered that style of approach but it does add extra logic to the helper, and does not seem as object-oriented as the store based approach. I also thought about the possible risk of ID collisions. Is this type of storage already implemented in some existing helper?

froschdesign:
> Do you mean something like `$this->attributes->add('someId', 'class', 'active')` and then `$this->attributes->toString('someId')`?

No no!

Instead of:

```php
$attributes = new AttributeStore(['class' => '…', 'title' => '…']);
echo $this->htmlAttributes($attributes);
echo $this->htmlAttributes($attributes);
```

this:

```php
$attributes = ['class' => '…', 'title' => '…'];
echo $this->htmlAttributes($attributes);
echo $this->htmlAttributes($attributes);
```

or:

```php
$attributes = new ArrayObject(['class' => '…', 'title' => '…']);
echo $this->htmlAttributes($attributes);
echo $this->htmlAttributes($attributes);
```

No need for extra utility classes.

aleksip:
That was basically the first version submitted in this PR. But dogfooding it I soon noticed that I am constantly repeating checks whether `class` is already set etc. in my templates.

So I find that the `set()`, `add()` and other methods in the utility class really help making the templates much more simple and the codebase more DRY.

Example:

```php
$attributes = $attributes ?? [];
$attributes['class'] = (array)($attributes['class'] ?? []);
$attributes['class'][] = 'nav';
$attributes['class'][] = 'navbar-nav';
echo $this->attributes($attributes);
```

with `AttributeStore`:

```php
$attributes = $this->attributes($attributes ?? []);
$attributes->addValue('class', ['nav', 'navbar-nav']);
echo $attributes;
```

There might be a more elegant way to do the first version, but the main point is that with the utility class this logic is in just one place, and additional features like checks can also be added at already the individual attribute stage, instead of only for the entire set in the end.

froschdesign:
> but the main point is that with the utility class this logic is in just one place

And my main point is: add this checks to the helper. No need for _utility_ class. 😄

aleksip:
I'm afraid I don't understand. In the examples you provided, the checking is done to the entire array? But it needs be done when adding attributes to the array. And if `$attributes = $this->attributes->add($attributes, 'class', 'active')` or `$this->attributes->add('someId', 'class', 'active')` are not the way to do it, what is?

aleksip:
`__construct()` and `setAttributes()` allow setting the entire array without checks in the current version of `AttributeStore`. This would need to be fixed.

froschdesign:
I have a few ideas, but I need some time for a short elaboration.

aleksip:
Ok, so meanwhile I have made an attempt to improve the class-based approach. `Attributes` is now an independent class extending from `ArrayObject`. I have removed the trait and instead refactored `AbstractHtmlElement` to use the new `Attributes` class.

I am not sure if two different escaper variables are actually required, but chose to be on the safe side. `AbstractHtmlElement::normalizeId()` would logically also belong to `Attributes`, but I have not moved it there yet because it needs to be overriden in some cases I am not too familiar with.

Based on a quick look at laminas-form, it might be possible to just "drop in" `Attributes` to `ElementInterface` elements, since it extends from `ArrayObject`. And then as a next step perhaps further develop and/or extend `Attributes` so that each element could have an element-specific `Attributes` configuration etc... I mention this just as a possibility, since even in their current form `Attributes` and `AttributesHelper` can be extremely useful.

src/Attributes.php
```diff
@@ -0,0 +1,142 @@
...
+     * @return Attributes
+     */
+    public function add($name, $value)
+    {
+        if ($this->offsetExists($name)) {
+            $this->offsetSet($name, array_merge((array) $this->offsetGet($name), (array) $value));
+        } else {
+            $this->offsetSet($name, $value);
+        }
+        return $this;
```
weierophinney:
```suggestion
        $this->offsetSet(
            $name,
             $this->offsetExists($name)
                ? array_merge((array) $this->offsetGet($name), (array) $value)
                : $value
        );
        return $this;
```

src/Attributes.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        if ($this->offsetExists($name)) {
+            $storeValue = $this->offsetGet($name);
+            if (is_array($storeValue)) {
+                return in_array($value, $storeValue);
+            } else {
+                return $value === $storeValue;
+            }
+        }
+        return false;
```
weierophinney:
```suggestion
        if (! $this->offsetExists($name)) {
            return false;
        }
        
        $storeValue = $this->offsetGet($name);
        if (is_array($storeValue)) {
            return in_array($value, $storeValue);
        }
        
        return $value === $storeValue;
```

src/Attributes.php
```diff
@@ -0,0 +1,142 @@
...
+            $val = $this->htmlAttrEscaper->escapeHtmlAttr($val);
+
+            if (strpos($val, '"') !== false) {
+                $xhtml .= " $key='$val'";
+            } else {
+                $xhtml .= " $key=\"$val\"";
+            }
+        }
+
+        return $xhtml;
```
weierophinney:
```suggestion
        $xhtml = '';

        foreach ($this->getArrayCopy() as $key => $val) {
            $key = $this->htmlEscaper->escapeHtml($key);

            if ((0 === strpos($key, 'on') || ('constraints' === $key)) && ! is_scalar($val)) {
                // Don't escape event attributes; _do_ substitute double quotes with singles
                // non-scalar data should be cast to JSON first
                $val = json_encode($val, JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP);
            }

            if (0 !== strpos($key, 'on') && 'constraints' !== $key && is_array($val)) {
                // Non-event keys and non-constraints keys with array values
                // should have values separated by whitespace
                $val = implode(' ', $val);
            }

            $val    = $this->htmlAttrEscaper->escapeHtmlAttr($val);
            $quote  = strpos($value, '"') !== false ? "'" : '"';
            $xhtml .= sprintf(' %2$s=%1$s%3$s%1$s', $quote, $key, $value);
        }

        return $xhtml;
```

Notes:

- laminas-json is in the process of being deprecated, as it doesn't really provide any benefits over ext-json
- Ideally, I'd extract the parts where you massage the `$val` into a method, so you could return early. However, two conditionals in sequence are better than nested conditionals.
- I use positional placeholders with `sprintf` to perform quoting (e.g. `%1$s` says to inject argument 1 as a string; I do this so we can re-use the `$quote` character we've previously discovered); variable interpolation is largely considered a no-no within Laminas components.

weierophinney (changes requested):
Looking good; I've suggested some improvements to make the various code paths more clear, and to reduce conditional nesting.

Ping me when you're ready for merge!

aleksip:
@weierophinney Thanks! I have now applied your suggestions.

Just to be sure: the class names and packages are fine, and there should be two variables for the escaper?

This is still a very minimal implementation and many useful features could be added, including features from the abstract view helper of laminas-form. More features could also be added in a subclass, of course. If this were to be merged in the current state, would it still be easy to add non-BC-breaking features? Or would it be better to "dogfood" this for a while longer?

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+    protected $htmlAttrEscaper;
+
+    /**
+     * Constructor.
+     *
+     * @param Escaper $htmlEscaper HTML escaper
+     * @param Escaper $htmlAttrEscaper Attribute escaper
+     * @param array|Traversable $attribs Attributes
+     */
+    public function __construct($htmlEscaper, $htmlAttrEscaper, $attribs = [])
```
weierophinney:
```suggestion
     * @param Escaper $htmlEscaper General HTML escaper
     * @param Escaper $htmlAttrEscaper Escaper for use with HTML attributes
     * @param iterable $attribs Attributes to manage
     */
    public function __construct(Escaper $htmlEscaper, Escaper $htmlAttrEscaper, iterable $attribs = [])
```

The 2.12.x sets the minimum supported PHP version at 7.3, so new features and classes, such as this one, can use the full range of features from that version.

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+     * Add a value to an attribute.
+     *
+     * Sets the attribute if it does not exist.
+     *
+     * @param $name string Name
+     * @param $value string|array Value
+     *
+     * @return Attributes
+     */
+    public function add($name, $value)
```
weierophinney:
```suggestion
     * @param $value string|array Value
     */
    public function add(string $name, $value): self
```

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Merge attributes with existing attributes.
+     *
+     * @param $attribs array|Traversable Attributes
+     *
+     * @return $this
+     */
+    public function merge($attribs)
```
weierophinney:
```suggestion
     */
    public function merge(iterable $attribs): self
```

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Does a specific attribute with a specific value exist?
+     *
+     * @param $name string Name
+     * @param $value string Value
+     *
+     * @return bool
+     */
+    public function hasValue($name, $value)
```
weierophinney:
```suggestion
     */
    public function hasValue(string $name, string $value): bool
```

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+        
+        return $value === $storeValue;
+    }
+
+    /**
+     * Return a string of tag attributes.
+     *
+     * @return string
+     */
+    public function __toString()
```
weierophinney:
```suggestion
     */
    public function __toString(): string
```

src/Helper/AttributesHelper.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Returns a new Attributes object, optionally initializing it with the
+     * provided value.
+     *
+     * @param array|Traversable $attribs Attributes
+     *
+     * @return Attributes
+     */
+    public function __invoke($attribs = [])
```
weierophinney:
```suggestion
     */
    public function __invoke(iterable $attribs = []): Attributes
```

weierophinney (changes requested):
This is really looking great, and I hate to provide more changes, but it's all in the interest of polishing it up and getting it ready for release.

As noted below: this branch requires a minimum supported PHP version of 7.3, so new functionality can make full use of parameter and return typehints, including scalar hints and the `iterable` pseudo-type.

Finally: please create a document in the `docs/book/helpers/` folder detailing the new helper and its usage; you can use one of the others in that directory as a template. When done, add it to the list in the `mkdocs.yml` file.

At that point, this will be ready to go!

src/Helper/AttributesHelper.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Laminas\View\Helper;
+
+use Laminas\View\Attributes;
+use Traversable;
+
+/**
+ * Helper for creating Attributes objects
+ */
+class AttributesHelper extends AbstractHelper
```
froschdesign: We have the following helpers:

* [HtmlTag](https://docs.laminas.dev/laminas-view/helpers/html-tag/)
* [HtmlList](https://docs.laminas.dev/laminas-view/helpers/html-list/)
* [HtmlObject](https://docs.laminas.dev/laminas-view/helpers/html-object/)
* …

And this helper creates HTML attributes therefore the name should also contain the prefix "Html".
With the [support of autocompletion](https://github.com/laminas/laminas-view/blob/c34b15f42cac410f35010163de169ab71c39bf73/src/Renderer/PhpRenderer.php#L25-L75) in IDEs, the length of the name is not a major problem.

src/Helper/AttributesHelper.php
froschdesign: No need to use abbreviations here, use `$attributes` instead of `$attribs`.

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+    {
+        $xhtml = '';
+
+        foreach ($this->getArrayCopy() as $key => $val) {
```
froschdesign: No need to use abbreviations here, use `$value` instead of `$val`.

src/Attributes.php
froschdesign: No need to use abbreviations here, use `$attributes` instead of `$attribs`.

src/Attributes.php
froschdesign: No need to use abbreviations here, use:

* `$htmlAttributeEscaper` instead of `$htmlAttrEscaper`
* `$attributes` instead of `$attribs`.

src/Attributes.php
```diff
@@ -0,0 +1,140 @@
...
+     * @var Escaper
+     */
+    protected $htmlEscaper;
+
+    /**
+     * HTML attribute escaper
+     *
+     * @var Escaper
+     */
+    protected $htmlAttrEscaper;
```
froschdesign: No need to use abbreviations here, use `$htmlAttributeEscaper` instead of `$htmlAttrEscaper`.

src/Helper/AttributesHelper.php
froschdesign: The new helper must be added to the DocBlock of [`Laminas\View\Renderer\PhpRenderer`](https://github.com/laminas/laminas-view/blob/c34b15f42cac410f35010163de169ab71c39bf73/src/Renderer/PhpRenderer.php#L75).

froschdesign (changes requested):
A few small changes are required. The most important is the name of the helper.

froschdesign:
@aleksip 
I can take over the extension of the documentation.

src/Helper/AttributesHelper.php
aleksip: @froschdesign Should `Attributes` also be renamed to `HtmlAttributes`? If not, maybe the `Helper` suffix could be dropped, since it is not used in other helpers apart from the abstract base class? I only added the suffix to make the helper class name distinct from the `Attributes` class.

froschdesign: @aleksip 
> …the Helper suffix could be dropped…

Correct. Use `HtmlAttributes` as class name for the helper and I think we can use `HtmlAttributesSet` for the current `Attributes` class.

aleksip:
Ok, I have hopefully now done everything that has been requested so far!

I also added `HtmlAttributesSet::set(iterable $attributes)`, which I have found useful, but it can of course be removed.

@froschdesign I was not sure what you meant by taking over the extension of the documentation. I have now added a very minimal initial version.

Since all the helper does is create new instances of `HtmlAttributesSet`, I was also not sure where to document the usage of that class.

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,17 @@
+# HtmlAttributes
+
+The `HtmlAttributes` helper is used to create `HtmlAttributesSet` objects.
```
froschdesign: > …to create `HtmlAttributesSet` objects.

This is not the goal of the helper and what the user wants. The true benefit and goal must be described here.
The class `Laminas\View\HtmlAttributesSet` should also be mentioned here and also that the helpers `EscapeHtml` and `EscapeHtmlAttr` are used – with link.

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Basic Usage
+
+```php
+$exampleAttributes = $this->htmlAttributes();
+$exampleAttributes->add('class', 'example');
+<div<?= $exampleAttributes ?>>
```
froschdesign: The example will not work because it mixes PHP and HTML without separation. And instead of `foo`, `bar` or `example`, a real example should be shown.

```suggestion
<?php
$attributes = $this->htmlAttributes(['class' => 'input-group']);
if ($form->hasValidated()) {
    $attributes->add('class', 'has-validation');
}
?>
<div<?= $attributes ?>></div>
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+Output:
+
+```html
+<div class="example">
```
froschdesign:
```suggestion
<div class="input-group&#x20;has-validation"></div>
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,17 @@
...
+```
```
froschdesign, line 148: The description of the class `Laminas\View\HtmlAttributesSet` is missing. All methods of the class (`set`, `add`, `merge`, `hasValue`, `__toString`) must be described and each with a code example. Also the fact that this class extends PHP's `ArrayObject` – with link to the related PHP documentation.

froschdesign (changes requested):
The user expects complete documentation, so further additions are still necessary.

Thanks in advance! 👍

docs/book/helpers/html-attributes.md
froschdesign: You can adapt the structure from the view helper pages of laminas-i18n. For example:

* https://docs.laminas.dev/laminas-i18n/view-helpers/currency-format/
* https://docs.laminas.dev/laminas-i18n/view-helpers/date-format/
* https://docs.laminas.dev/laminas-i18n/view-helpers/number-format/

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
+# HtmlAttributes
+
+The `HtmlAttributes` helper can be used to make code related to HTML attributes
+easier to write and read. The helper initializes and returns
```
froschdesign: Words like "easy" and "simple" should be avoid in a documentation.

1. A documentation is not an advertisement.
1. Just because it's easier for you or me doesn't make it easier for others.
1. And should it not work as described, then these words lead to frustration.

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+<div<?= $attributes ?>></div>
+```
+
+Output:
+
+```html
+<div class="input-group&#x20;has-validation"></div>
+```
+
+## Getting an `HtmlAttributesSet` object instance
```
froschdesign: For headlines we use the Chicago Manual of Style capitalization rules:
```suggestion
## Getting an `HtmlAttributesSet` Object Instance
```
This tool can help here: https://capitalizemytitle.com/style/Chicago/

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+$attributes = $this->htmlAttributes([
+    'id' => 'login-username',
+    'class' => ['input-group', 'mb-3']
+]);
+```
+
+Calling the helper always creates a new object instance. Several
+`HtmlAttributesSet` object instances can be used in the same template.
+
+## Using `HtmlAttributesSet` as an array
```
froschdesign:
```suggestion
## Using `HtmlAttributesSet` as an Array
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+
+`HtmlAttributeSet` extends PHP's
+[`ArrayObject`](https://www.php.net/manual/en/class.arrayobject.php) which
+allows it to be used like an array.
+
+### Setting an attribute
```
froschdesign:
```suggestion
### Setting an Attribute
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+
+```php
+$attributes['id'] = 'login-username';
+
+$attributes['class'] = ['input-group', 'mb-3'];
+```
+
+## Setting several attributes at once
```
froschdesign:
```suggestion
## Setting Several Attributes at Once
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+`HtmlAttributesSet::set(iterable $attributes)` method.
+
+```php
+$attributes->set([
+    'id' => 'login-username',
+    'class' => ['input-group', 'mb-3']
+])
+```
+
+## Adding a value to an attribute
```
froschdesign:
```suggestion
## Adding a Value to an Attribute
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+
+Output:
+
+```html
+<div class="input-group"></div>
+
+<div class="input-group&#x20;has-validation"></div>
+```
+
+## Merging attributes with existing attributes
```
froschdesign:
```suggestion
## Merging Attributes with Existing Attributes
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+<div<?= $attributes ?>></div>
+```
+
+Output:
+
+```html
+<div id="login-username" class="input-group&#x20;mb-3"></div>
+```
+
+## Checking if a specific attribute with a specific value exists
```
froschdesign:
```suggestion
## Checking If a Specific Attribute with a Specific Value Exists
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+
+The method handles cases where the attribute does not exist or has multiple values.
+
+```php
+if ($attributes->hasValue('class', 'has-validation')) {
+    // ...
+}
+```
+
+## Outputting attributes
```
froschdesign:
```suggestion
## Outputting Attributes
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+Output:
+
+```html
+<div class="input-group&#x20;has-validation"></div>
+```
+
+## Getting an `HtmlAttributesSet` object instance
+
+To get an empty `HtmlAttributesSet` object instance, call the helper without any
+parameters.
```
froschdesign: Documentation should be written using the [one sentence per-line method](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line).

```suggestion
To get an empty `HtmlAttributesSet` object instance, call the helper without any parameters.
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+When an `HtmlAttributesSet` object is converted to a string, attribute names
+and values are automatically escaped using escapers from the
+[EscapeHtml](https://docs.laminas.dev/laminas-view/helpers/escape/#escapehtml)
+and [EscapeHtmlAttr](https://docs.laminas.dev/laminas-view/helpers/escape/#escapehtmlattr)
+view helpers.
+
+```php
+<?php
+$attributes = $this->htmlAttributes([
+    'title' = 'faketitle onmouseover=alert(/laminas-framework/);'
```
froschdesign:
```suggestion
    'title' = 'faketitle onmouseover=alert(/laminas-project/);'
```

docs/book/helpers/html-attributes.md
```diff
@@ -0,0 +1,161 @@
...
+]);
+?>
+
+<a<?= $attributes ?>>click</a>
+```
+
+Output:
+
+```html
+<a title="faketitle&#x20;onmouseover&#x3D;alert&#x28;&#x2F;laminas-framework&#x2F;&#x29;&#x3B;">click</a>
```
froschdesign:
```suggestion
<a title="faketitle&#x20;onmouseover&#x3D;alert&#x28;&#x2F;laminas-project&#x2F;&#x29;&#x3B;">click</a>
```

froschdesign (changes requested):
Some changes for the documentation are required.

Thanks in advance! 👍

froschdesign (approved):

froschdesign:
@weierophinney 
Something is wrong here. The last version is 2.12.0 but the branch 2.13.x for the next minor version is missing?! Should this not be created by the automatic workflow?
The milestone for 2.13 exists.

weierophinney:
> Something is wrong here. The last version is 2.12.0 but the branch 2.13.x for the next minor version is missing?! Should this not be created by the automatic workflow?
> The milestone for 2.13 exists.

Check the logs for the 2.12.0 release to see if any errors were reported, and report that against the automatic-releases repo. In the meantime, you can create and push it manually based on the ref used for the 2.12.0 release.

aleksip:
@weierophinney @froschdesign Hi, I noticed that this PR is still listed with "Changes requested" and is "In Progress" in the Minor improvements for components project. Is there anything left I can/should do?

froschdesign:
@aleksip 
Your new feature will be released with the [next minor version](https://github.com/laminas/laminas-view/milestone/4) but before that a new branch (2.13.x) is needed and then all related pull requests must use it.

At the moment we are working on the following projects:

* [Switch from Travis-Ci to GitHub Actions](https://github.com/orgs/laminas/projects/14)
* [PHP 8.0 support](https://github.com/orgs/laminas/projects/11)

I will try to find time to review the other pull requests for 2.13.

Thanks for your patience! 👍

aleksip:
@froschdesign Thanks for your quick reply! No worries regarding release date, just wanted to make sure that I have not missed anything and that the PR will not be skipped from the next release because of the "Changes requested" and "In Progress" states.

froschdesign:
@aleksip 
Thanks for your understanding!

I will inform you when the target branch is available. In the meantime, maybe you can take a look at the open issue reports, some more help is needed. Thanks in advance! 😃

weierophinney (approved):

PR #63, opened 2021-02-25
Remove `laminas/laminas-config` dependency

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

With reference to [TSC issue #68](https://github.com/laminas/technical-steering-committee/issues/68), this pull removes the direct dependency on `laminas/laminas-config`


weierophinney:
Since this changes dependencies, the patch should target next minor; I've updated it to target the 2.13.0 milestone. You'll need to rebase, I think.

weierophinney:
This is going to be a bit harder than replacing `Laminas\Config\Config` with `ArrayObject`. The navigation helper tests use `Laminas\Config\Factory::fromFile()` with **an XML config file** to seed the test cases. We would need to redo the assets those tests use as pure PHP, JSON, or by writing a quick XML parser. As such, I'm going to remove the milestone from this for now.

gsteel:
> This is going to be a bit harder than replacing `Laminas\Config\Config` with `ArrayObject`. The navigation helper tests use `Laminas\Config\Factory::fromFile()` with **an XML config file** to seed the test cases. We would need to redo the assets those tests use as pure PHP, JSON, or by writing a quick XML parser. As such, I'm going to remove the milestone from this for now.

I've just been looking at these - not sure how I didn't spot that in the first place. Is it worth considering splitting the navigation related helpers into their own package?

froschdesign:
@gsteel 
You can convert the navigation configuration into a PHP array. laminas-config can help here. 😄

test/Helper/Navigation/LinksTest.php
```diff
@@ -167,7 +167,7 @@ public function testDetectRelationFromArrayPropertyOfActivePage(): void
     public function testDetectRelationFromConfigInstancePropertyOfActivePage(): void
```
froschdesign: The method name for this test should be updated as "Config" is no longer used.

test/Helper/Navigation/LinksTest.php
```diff
@@ -214,7 +214,7 @@ public function testDetectMultipleRelationsFromConfigPropertyOfActivePage(): voi
     {
         $active = $this->_helper->findOneByLabel('Page 2');
 
-        $active->addRel('alternate', new Config\Config([
+        $active->addRel('alternate', new ArrayObject([
```
froschdesign, line 213: The method name for this test should be updated as "Config" is no longer used.

gsteel: > @gsteel
> You can convert the navigation configuration into a PHP array. laminas-config can help here. 😄

Ha ha!!! Yes - that's what I'm using now!

gsteel:
I haven't updated the lock file or changed the target to 2.14 - it's currently even with 2.13 and the lock file is likely to cause merge conflicts if anything else gets merged. I also haven't updated psalm baseline ~200 errors have been fixed.

Ocramius (approved):
:ship: awesome, thanks @gsteel!

composer.json
```diff
@@ -25,10 +25,10 @@
         "laminas/laminas-stdlib": "^3.6"
     },
     "require-dev": {
+        "ext-dom": "*",
         "laminas/laminas-authentication": "^2.5",
         "laminas/laminas-cache": "^2.6.1",
         "laminas/laminas-coding-standard": "~1.0.0",
-        "laminas/laminas-config": "^3.7",
```
Ocramius, line 31: Oh, this change does indeed require a `composer.lock` change - I'll see if I can do it...

gsteel: I can update the lock now if it suits - I have 7.3 handy

Ocramius: @gsteel yes please: a `composer update laminas/laminas-config` should suffice to keep the `composer.lock` diff minimal

Ocramius: Cool, `laminas/laminas-config` is done from here as a direct dependency! 

Still in there, but the `composer.lock` update is correct, so far :-)

```
root@f5e835d0a0ab:/app# composer why laminas/laminas-config
laminas/laminas-modulemanager  2.10.2  requires  laminas/laminas-config (^3.4) 
```

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #64, opened 2021-02-25
Fix tests without assertions

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

A number of the tests _(Around 30)_ complain about no assertions which is pretty noisy when you're running tests locally.

This pull updates those tests to use `$this->expectedException()` where appropriate instead of try/fail and add `$this->addToAssertionCount()` when appropriate.

There's also a little bit of de-duplication surrounding the escape helpers and permitted encodings and updates to mock usage so that the mocks assert some expectations where previously there were none.

### DIFF
--- a/test/Helper/EscaperEncodingsTrait.php
+++ b/test/Helper/EscaperEncodingsTrait.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\View\Helper;
+
+trait EscaperEncodingsTrait
+{
+    /** @var string[] */
+    private $supportedEncodings = [
+        'iso-8859-1',   'iso8859-1',    'iso-8859-5',   'iso8859-5',
+        'iso-8859-15',  'iso8859-15',   'utf-8',        'cp866',
+        'ibm866',       '866',          'cp1251',       'windows-1251',
+        'win-1251',     '1251',         'cp1252',       'windows-1252',
+        '1252',         'koi8-r',       'koi8-ru',      'koi8r',
+        'big5',         '950',          'gb2312',       '936',
+        'big5-hkscs',   'shift_jis',    'sjis',         'sjis-win',
+        'cp932',        '932',          'euc-jp',       'eucjp',
+        'eucjp-win',    'macroman'
+    ];
+
+    /** @return iterable<string, array<int, string>> */
+    public function supportedEncodingsProvider() : iterable
+    {
+        foreach ($this->supportedEncodings as $encoding) {
+            yield $encoding => [$encoding];
+        }
+    }
+}
--- a/test/Helper/EscaperEncodingsTrait.php
+++ b/test/Helper/EscaperEncodingsTrait.php
@@ -1,5 +1,11 @@
 <?php
 
+/**
+ * @see       https://github.com/laminas/laminas-view for the canonical source repository
+ * @copyright https://github.com/laminas/laminas-view/blob/master/COPYRIGHT.md
+ * @license   https://github.com/laminas/laminas-view/blob/master/LICENSE.md New BSD License
+ */
+
 declare(strict_types=1);
 
 namespace LaminasTest\View\Helper;
@@ -20,7 +26,7 @@ trait EscaperEncodingsTrait
     ];
 
     /** @return iterable<string, array<int, string>> */
-    public function supportedEncodingsProvider() : iterable
+    public function supportedEncodingsProvider(): iterable
     {
         foreach ($this->supportedEncodings as $encoding) {
             yield $encoding => [$encoding];
--- a/test/Helper/HeadLinkTest.php
+++ b/test/Helper/HeadLinkTest.php
@@ -321,7 +321,7 @@ class HeadLinkTest extends TestCase
         $this->assertStringContainsString('<!--<![endif]-->', $string);
     }
 
-    public function argumentCountProvider() : iterable
+    public function argumentCountProvider(): iterable
     {
         return [
             'One' => [1],
--- a/test/Helper/HeadLinkTest.php
+++ b/test/Helper/HeadLinkTest.php
@@ -9,10 +9,12 @@
 namespace LaminasTest\View\Helper;
 
 use Laminas\View\Exception;
-use Laminas\View\Exception\ExceptionInterface as ViewException;
 use Laminas\View\Helper;
 use Laminas\View\Renderer\PhpRenderer as View;
 use PHPUnit\Framework\TestCase;
+
+use function array_fill;
+use function sprintf;
 
 /**
  * Test class for Laminas\View\Helper\HeadLink.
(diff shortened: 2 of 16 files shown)


test/Helper/EscaperEncodingsTrait.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
```
froschdesign: Please add the licence header for this file.

test/Helper/HeadLinkTest.php
```diff
@@ -319,18 +321,24 @@ public function testConditionalStylesheetCreationNoIEWidthSpaces()
         $this->assertStringContainsString('<!--<![endif]-->', $string);
     }
 
-    public function testSettingAlternateWithTooFewArgsRaisesException()
+    public function argumentCountProvider() : iterable
```
froschdesign: Please run the coding standard check. There should be no space before the colon.

test/Helper/HeadLinkTest.php
```diff
@@ -319,18 +321,24 @@ public function testConditionalStylesheetCreationNoIEWidthSpaces()
...
-            $this->fail('Setting alternate with fewer than 3 args should raise exception');
-        } catch (ViewException $e) {
-        }
+        return [
+            'One' => [1],
+            'Two' => [2],
+        ];
+    }
+
+    /** @dataProvider argumentCountProvider */
```
froschdesign, line 332: Please run the coding standard check. I don't know if the short DocBlock is allowed for methods.

gsteel:
Whoops… Just rebased on 2.12.x - was I wrong to do that?

test/Helper/HeadMetaTest.php
```diff
@@ -77,27 +77,34 @@ public function testHeadMetaReturnsObjectInstance()
...
     }
 
-    public function testAppendPrependAndSetThrowExceptionsWhenNonMetaValueProvided()
+    public function testAppendPrependAndSetThrowExceptionsWhenNonMetaValueProvided(): void
     {
         try {
             $this->helper->append('foo');
             $this->fail('Non-meta value should not append');
         } catch (ViewException $e) {
+            $this->addToAssertionCount(1);
```
froschdesign: Please test here that an exception is thrown.

test/Helper/HeadScriptTest.php
```diff
@@ -72,21 +72,28 @@ public function testSetPrependAppendAndOffsetSetThrowExceptionsOnInvalidItems()
             $this->helper->append('foo');
             $this->fail('Append should throw exception with invalid item');
         } catch (View\Exception\ExceptionInterface $e) {
+            $this->addToAssertionCount(1);
```
froschdesign: The same here, test that an exception is thrown.

test/Helper/Navigation/BreadcrumbsTest.php
```diff
@@ -233,17 +233,14 @@ public function testRenderingPartialBySpecifyingAnArrayAsPartial()
...
-            $this->fail(
-                '$partial was invalid, but no Laminas\View\Exception\ExceptionInterface was thrown'
-            );
-        } catch (ExceptionInterface $e) {
-        }
+        $this->expectException(InvalidArgumentException::class);
+        $this->_helper->render();
+        $this->fail(
+            '$partial was invalid, but no Laminas\View\Exception\ExceptionInterface was thrown'
+        );
```
froschdesign: I think calling the fail is not needed anymore.

test/Helper/PartialLoopTest.php
```diff
@@ -227,13 +226,14 @@ public function testShouldNotCastToArrayIfObjectIsTraversable()
...
      */
-    public function testEmptyArrayPassedToPartialLoopShouldNotThrowException()
+    public function testEmptyArrayPassedToPartialLoopShouldNotThrowException(): void
     {
         $view = new View();
         $view->resolver()->addPath($this->basePath . '/application/views/scripts');
         $this->helper->setView($view);
 
         $this->helper->__invoke('partialLoop.phtml', []);
+        $this->addToAssertionCount(1);
```
froschdesign: I think the partial counter can be compared here.

froschdesign (changes requested):
Thanks for your contribution! 😃 

I added some suggestions for updates. Thanks in advance! 👍

froschdesign:
@gsteel 
> Whoops… Just rebased on 2.12.x - was I wrong to do that?

This will be shipped as a bugfix so the branch for latest version is correct.

test/Helper/HeadLinkTest.php
gsteel: CS checks pass - you're right of course, but for some reason it's not picked up. Coding standard 1.x is in use - perhaps another pull to update it to ~2 at some point?

test/Helper/Navigation/BreadcrumbsTest.php
gsteel: No… it's not needed but I thought leaving it would aid docs ?

froschdesign: The call of `expectException` is enough already.

test/Helper/HeadLinkTest.php
froschdesign: The new continuous integration workflow must be added before. A new pull request needed like this one: https://github.com/laminas/laminas-paginator/pull/18

test/Helper/HeadLinkTest.php
gsteel: Just checked this with coding-std ~2 and the short doc block is preferred. ~1 doesn't complain.

test/Helper/HeadScriptTest.php
```diff
@@ -216,27 +220,27 @@ public function testOverloadOffsetSetScriptWritesToSpecifiedIndex()
...
     }
 
-    public function testOverloadingWithTooFewArgumentsRaisesException()
+    public function testOverloadingWithTooFewArgumentsRaisesException(): void
     {
         try {
             $this->helper->setScript();
             $this->fail('Too few arguments should raise exception');
         } catch (View\Exception\ExceptionInterface $e) {
+            $this->addToAssertionCount(1);
```
froschdesign: Please remove all `addToAssertionCount` as used here.

test/Helper/Navigation/MenuTest.php
```diff
@@ -294,15 +295,12 @@ public function testRenderingPartialWithParams()
...
         $this->_helper->setPartial(['menu.phtml']);
-
-        try {
-            $this->_helper->render();
-            $this->fail('invalid $partial should throw Laminas\View\Exception\InvalidArgumentException');
-        } catch (\Laminas\View\Exception\ExceptionInterface $e) {
-        }
+        $this->expectException(InvalidArgumentException::class);
+        $this->_helper->render();
+        $this->fail('invalid $partial should throw Laminas\View\Exception\InvalidArgumentException');
```
froschdesign: Please remove also all `fail`.

Ocramius (approved):
LGTM! Will aggregate related patches together, so we can see if it all goes green after merge :)

PR #70, opened 2021-03-01
Update Composer lock file and fix test failures

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Updates the composer lock file with i18n-mvc and flash messenger packages.

* Tests should still fail on lowest because of the namespace rename from `Laminas\Mvc\Router` to `Laminas\Router`
* The `UrlIntegrationTest` should still fail because the request retrieved from the container is not a `ConsoleRequest`

### DIFF
--- a/test/Helper/Navigation/AbstractTest.php
+++ b/test/Helper/Navigation/AbstractTest.php
@@ -10,7 +10,6 @@ namespace LaminasTest\View\Helper\Navigation;
 
 use Laminas\Config\Factory as ConfigFactory;
 use Laminas\I18n\Translator\Translator;
-use Laminas\Mvc\Router\RouteMatch as V2RouteMatch;
 use Laminas\Mvc\Service\ServiceManagerConfig;
 use Laminas\Navigation\Navigation;
 use Laminas\Navigation\Service\DefaultNavigationFactory;
@@ -18,7 +17,7 @@ use Laminas\Permissions\Acl\Acl;
 use Laminas\Permissions\Acl\Resource\GenericResource;
 use Laminas\Permissions\Acl\Role\GenericRole;
 use Laminas\Router\ConfigProvider as RouterConfigProvider;
-use Laminas\Router\RouteMatch as V3RouteMatch;
+use Laminas\Router\RouteMatch;
 use Laminas\ServiceManager\Config;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\View\Renderer\PhpRenderer;
@@ -77,8 +76,6 @@ abstract class AbstractTest extends TestCase
      * @var Navigation\Navigation
      */
     protected $_nav3;
-
-    private $_oldControllerDir;
     // @codingStandardsIgnoreEnd
 
     /**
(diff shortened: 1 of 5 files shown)


gsteel:
The UrlIntegrationTest is problematic in that it is unclear to me why a legacy console request is expected. I believe that in the past we may have been able to guarantee a console request was present. Given that Console itself is deprecated, can this test be left in a flaky state, given that the expected output is still the same? I would imagine that the test will be deleted in the not too distant future anyhow… Paging @Ocramius for review :)

Ocramius:
> it is unclear to me why a legacy console request is expected

Probably for testing purposes - the test is most likely there to ensure BC compatibility.

 > Given that Console itself is deprecated, can this test be left in a flaky state, given that the expected output is still the same?

Sadly not: deprecation is not removal of support.

 >  I would imagine that the test will be deleted in the not too distant future anyhow…

On the `3.0.0` release, we can certainly remove it, but in `2.x`, it must stay in place if it was ever green in the past.

test/Helper/Navigation/AbstractTest.php
```diff
@@ -89,10 +86,6 @@ protected function setUp(): void
     {
         $cwd = __DIR__;
 
-        $this->routeMatchType = class_exists(V2RouteMatch::class)
```
Ocramius, line 92: This conditional indicates that we can deal with V2 **or** V3. Can we somehow reduce that via `composer.json` restrictions (support only one version)?

test/Helper/UrlIntegrationTest.php
```diff
@@ -29,16 +27,15 @@
  */
 class UrlIntegrationTest extends TestCase
 {
+    private $serviceManager;
+
     protected function setUp(): void
     {
-        $this->literalRouteType = class_exists(V2HttpRoute\Literal::class)
```
Ocramius, line 34: Similar to above: putting a restriction in `composer.json` is necessary for this conditional to be removed.

test/Helper/UrlTest.php
```diff
@@ -52,39 +46,15 @@ class UrlTest extends TestCase
      */
     protected function setUp(): void
     {
-        $this->routeMatchType = class_exists(RouteMatch::class)
```
Ocramius, line 55: Similar discussion here: `Laminas\Router` dependency must be bumped

Ocramius (changes requested):

test/Helper/Navigation/AbstractTest.php
gsteel: I've made the minimum version of Laminas\Navigation ^2.8.1 in dev which is the version where support was added for Laminas\Router vs Laminas\Mvc\Router.

Ocramius: Does `"laminas/laminas-router": "^3.0.1",` cover that then?

gsteel: I assumed that if --dev will only install v3, then it's simply not possible to test against v2 without changing the dev requirement of router to 2||3 or something like that - so 2.8.1 of navigation fixes the test failures and renders the checks for v2 router obsolete???

gsteel: P.S. I think that the dev requirement of navigation could actually be bumped 2.9.x due to 7.3 support getting added around that release

Ocramius: We can add a `"conflict": {"laminas/laminas-router": "<3.0.1"}` to solve that

Ocramius:
This was manually merged, thanks @gsteel!

PR #75, opened 2021-05-10
Remove file headers

ghostwriter:
**[QA]** Remove file header (DocBlock comments)


Ocramius (approved):
:ship: 

I did a quick rebase, but it's a net improvement at zero cost :)

Issue #71, opened 2021-03-01
`2.12.x` branch to be renamed `2.13.x`, `2.12.x` to be re-created from `2.12.0`


Ocramius:
The current `2.12.x` branch is effectively `2.13.x`, and should **not** be released as-is.

TODOs:

 * [ ] rename `2.12.x` to `2.13.x`
 * [ ] fix `2.13.x` build (fails on markdown linting)
 * [ ] re-create `2.12.x` from `2.12.0` tagged release
 * [ ] move issues from milestone `2.12.1` to milestone `2.13.0`

gsteel:
#72 Should solve the markdown linting problem but may well require review from @froschdesign … I've setup some linting rules to help the docs pass with as little modification to the existing docs as possible so there's not too much noise and those rules may or may not be OK

gsteel:
Perhaps we can get 2.13.x out the door if I re-open _(and fix)_  #72 - Maybe I could ask for a bit of help from @froschdesign or @weierophinney on the docs formatting??? That would get CI working nice and allow other pulls to be pushed thru :D

froschdesign:
@gsteel 
You can re-open #72 but all changes to the line lengths must be revert. The configuration for the markdownlint is already [updated](https://github.com/laminas/laminas-continuous-integration-action/blob/bf7d42f3d8eb1a0b817ded98db13cfc899f4440e/etc/markdownlint.json#L3).
The special Markdown syntax for [tabs](https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/) can be excluded from the linting via the [HTML comments](https://github.com/DavidAnson/markdownlint#configuration).

weierophinney:
I'm a bit confused at this point. I think we can go forward with 2.12.1 from the current state of 2.12.x, as it has 7 PRs already merged against it, all since 2.12.0 was released, and that will then create the 2.13.x branch. We can then re-target existing PRs that we've tagged for 2.13.0 against that.

PR #77, opened 2021-07-08
Add support for the nonce script attribute

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| New Feature   | yes

### Description

The [nonce attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) is useful for implementing a content security policy. Whilst it is possible to allow arbitrary attributes for the helper(s), in order to add a nonce to a <script> you'd have to implement your own delegator or factory to set this flag.

### DIFF
--- a/test/Helper/HeadScriptTest.php
+++ b/test/Helper/HeadScriptTest.php
@@ -551,4 +551,18 @@ document.write(bar.strlen());');
         $test = $this->helper->__invoke()->toString();
         $this->assertDoesNotMatchRegularExpression('#type="text/javascript"#i', $test);
     }
+
+    public function testSupportsNonceAttribute(): void
+    {
+        ($this->helper)()->appendScript(
+            '// some js',
+            'text/javascript',
+            ['nonce' => 'random']
+        );
+
+        self::assertStringContainsString(
+            'nonce="random"',
+            (string) ($this->helper)()
+        );
+    }
 }
--- a/test/Helper/HeadScriptTest.php
+++ b/test/Helper/HeadScriptTest.php
@@ -8,9 +8,12 @@
 
 namespace LaminasTest\View\Helper;
 
+use Generator;
 use Laminas\View;
 use Laminas\View\Helper;
 use PHPUnit\Framework\TestCase;
+
+use function sprintf;
 
 /**
  * Test class for Laminas\View\Helper\HeadScript.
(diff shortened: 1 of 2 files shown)


gsteel:
PHPCS failure due to being locked at 2.x by laminas coding standard 1.x - phpcs won't pass until everything is updated to use coding std 2.x

Also, targeting 2.12 because of #71

froschdesign:
@gsteel 
I think all the other missing attributes can be added as well: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attributes

test/Helper/HeadScriptTest.php
```diff
@@ -551,4 +554,43 @@ public function testOmitsTypeAttributeIfNoneGivenAndHtml5Doctype()
...
+    {
+        $values = ['async', 'defer', 'nomodule'];
+
+        foreach ($values as $name) {
+            yield $name => [$name];
+        }
+    }
+
+    /** @dataProvider booleanAttributeDataProvider */
+    public function testBooleanAttributesUseTheKeyNameAsTheValue(string $attribute): void
```
froschdesign, line 583: If you add the support for attributes that behave as [booleans](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute) then this should work as well, based on the doctype:

```html
<script type="module" async></script>
```

froschdesign (changes requested):
_Sorry for the very late response, I am busy with work at the moment._

Looks good so far but one thing is missing.
Thanks in advance! 👍

test/Helper/HeadScriptTest.php
gsteel: Hey @froschdesign - I'm not sure what you mean here. I've added a test to prove that boolean attributes can be assigned to script elements marked as modules. The style of the booleans is `attr="attr"` as opposed to `attr` - both are valid - are you suggesting to change the style to the latter?

Figuring out whether `nomodule` is mutually exclusive with a `module` type - do you mean that??

froschdesign: > The style of the booleans is `attr="attr"` as opposed to `attr` - both are valid…

Right, but with standard HTML5 (not XHTML5) the short version should be used or the user should be able to decide for himself.

But at this point we leave everything as it is, because with version 3 at the latest, the entire attribute creation must be revised. I don't like at all that the [abstract view helper in laminas-form](https://github.com/laminas/laminas-form/blob/c6a69c93e9c445b9a2e08cd7a4746bc724bd4ffe/src/View/Helper/AbstractHelper.php#L30) does something more or different.
And with #60 is something new in the pipeline.

froschdesign (approved):

PR #78, opened 2021-07-08
Fix Markdown Formatting for CI Round 2

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

With reference to #71 and #72, this pull solves formatting errors in Markdown files under the `docs` tree, mostly by using comments for markdownlint.

Calling on @froschdesign for review

PR #81, opened 2021-10-12
PHP8.1 support added

konarshankar07:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
PHP8.1 support added

Related ticket: https://github.com/magento/magento2/issues/34214

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,0 +1,5 @@
+{
+    "ignore_php_platform_requirements": {
+        "8.1": true
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -80,6 +80,7 @@
     },
     "autoload-dev": {
         "files": [
+            "polyfill/autoload.php",
             "test/autoload.php"
         ],
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -20,22 +20,21 @@
     },
     "require": {
         "php": "^7.3 || ~8.0.0 || ~8.1.0",
-        "laminas/laminas-eventmanager": "^3.0",
-        "laminas/laminas-json": "^2.6.1 || ^3.0",
-        "laminas/laminas-loader": "^2.5",
-        "laminas/laminas-stdlib": "^3.2.1",
-        "laminas/laminas-zendframework-bridge": "^1.0"
+        "laminas/laminas-eventmanager": "^3.4",
+        "laminas/laminas-json": "^2.6.1 || ^3.3",
+        "laminas/laminas-loader": "^2.8",
+        "laminas/laminas-stdlib": "^3.6"
     },
     "require-dev": {
         "laminas/laminas-authentication": "^2.5",
         "laminas/laminas-cache": "^2.6.1",
         "laminas/laminas-coding-standard": "~1.0.0",
-        "laminas/laminas-config": "^2.6",
+        "laminas/laminas-config": "^3.7",
         "laminas/laminas-console": "^2.6",
         "laminas/laminas-escaper": "^2.5",
-        "laminas/laminas-feed": "^2.7",
+        "laminas/laminas-feed": "^2.15",
         "laminas/laminas-filter": "^2.6.1",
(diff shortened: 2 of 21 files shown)


weierophinney:
@konarshankar07 I've added a few commits to get this working on latest deps and with 8.1; thanks for getting this started!

froschdesign:
@weierophinney 
There is a separate pull request for the Markdown problems: #78

weierophinney:
> There is a separate pull request for the Markdown problems: #78

I've just pushed a fix for them here, as I'd already started before you commented. :-/

Not entirely sure why it ran those checks, TBH - there were no changes to the docs, so those should not have run.

fascinosum:
hi @weierophinney, thank you and @konarshankar07 for the implemented support of PHP 8.1

@weierophinney, could you explain the reason for https://github.com/laminas/laminas-view/blob/2b396adcd8489826307b36427ed46c61e460e4e0/composer.json#L58
If project has `laminas/laminas-zendframework-bridge` as dependency, it does not allow this version to be installed, because `conflicts with zendframework/zend-router * (laminas/laminas-router 3.4.5 replaces zendframework/zend-router ^3.3.0)` 
You can check it with the latest available version of `laminas/laminas-mvc`

weierophinney:
> If project has `laminas/laminas-zendframework-bridge` as dependency, it does not allow this version to be installed, because `conflicts with zendframework/zend-router * (laminas/laminas-router 3.4.5 replaces zendframework/zend-router ^3.3.0)`
> You can check it with the latest available version of `laminas/laminas-mvc`

Oof - that was an incorrect line - it should be zend-view. I'll get that updated and issue a patch release.

fascinosum:
thank you, @weierophinney

konarshankar07 (commented):
@weierophinney 
Regarding https://github.com/laminas/laminas-view/pull/81/commits/6c73cf800bd08facab9f696df05cf819c6f85d98 
Thanks for finishing the pending work but can you please explain why we need to add the polyfill classes and how you are able to reproduce the ReturnTypeWillChange error?
I'm new to the laminas so curious about this commit

froschdesign:
@konarshankar07 
> …why we need to add the polyfill classes…

Components like laminas-filter and laminas-navigation do not yet support PHP 8.1, so the polyfills were created. Otherwise we have to wait for 3 other components and the related updates.

> …and how you are able to reproduce the ReturnTypeWillChange error?

Were you able to try [the hint from the comments in laminas-crypt](https://github.com/laminas/laminas-crypt/pull/14#issuecomment-941181297)?

konarshankar07:
@froschdesign 
> Were you able to try the hint from the comments in laminas-crypt?


Yes, I found that the test runs are executing with the php8.1.0RC4 and in my local I'm running the test using php8.1.0RC2 version

froschdesign:
@konarshankar07 
Right: https://github.com/laminas/laminas-view/runs/3873270121?check_suite_focus=true#step:3:18

Were you able to reproduce the errors in your local test environment?

konarshankar07:
@froschdesign 

> Were you able to reproduce the errors in your local test environment?

Nope, I'm not seeing any error in my local with the php8.1.0RC2

weierophinney:
@konarshankar07 

The trick is adding the `convertDeprecationsToExceptions="true"` configuration to the `phpunit.xml.dist` file.

Prior to PHPUnit 9.5.10, this was the default, which meant that any deprecations from the engine or other code would trigger an exception, which, unless you had an `expectDeprecation()` directive in the test, would cause it to error, and thus cause CI to fail.

With 9.5.10, PHPUnit decided to toggle that off.

We actually want it ON, because it allows us to identify deprecations we are relying on in our libraries _early_, and adapt the code to work for future versions. That was the change I made that caused tests to fail, and then required changes so that we could complete the patch.

PR #87, opened 2021-10-13
Fixed segfault in `htmlAttributes` helper used in `AbstractHtmlElement`, due to incorrect helper name reference

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| QA            | yes

### Description

With the introduction of `$this->htmlAttributes()` a kind-of BC break was introduced… For ages, I've been implementing `HtmlAttributes` helpers in my projects to basically expose the method in `AbstractHtmlElement::htmlAttribs()` which is a protected method. Because, the htmlAttribs() method was updated to call the plugin by alias `htmlAttributes`, this would cause an infinite loop and segfault in consumer code with an existing helper aliased to `htmlAttributes`.

The fix here just uses the FQCN of the expected helper instead of the alias.

I also added a quick test case for the helper itself whilst I was hunting down the cause of the segfault in my own code.

Also - the introduced [HtmlAttributesSet](https://github.com/laminas/laminas-view/blob/3.0.x/src/HtmlAttributesSet.php) class is untested and has several public methods - would it be worth doing some QA on that?

### DIFF
--- a/src/Helper/AbstractHtmlElement.php
+++ b/src/Helper/AbstractHtmlElement.php
@@ -64,7 +64,7 @@ abstract class AbstractHtmlElement extends AbstractHelper
             }
         }
 
-        $attribs = $this->getView()->plugin('htmlAttributes')($attribs);
+        $attribs = $this->getView()->plugin(HtmlAttributes::class)($attribs);
 
         return (string) $attribs;
     }
--- a/test/Helper/HtmlAttributesTest.php
+++ b/test/Helper/HtmlAttributesTest.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\View\Helper;
+
+use Laminas\View\Helper\HtmlAttributes;
+use Laminas\View\Renderer\PhpRenderer;
+use PHPUnit\Framework\TestCase;
+
+class HtmlAttributesTest extends TestCase
+{
+    /** @var HtmlAttributes */
+    private $helper;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->helper = new HtmlAttributes();
+        $this->helper->setView(new PhpRenderer());
+    }
+
+    public function testThatInvokeWillReturnAttributeSetWithTheExpectedValues(): void
+    {
+        $set = ($this->helper)([
+            'some' => 'value'
+        ]);
+
+        self::assertEquals(['some' => 'value'], $set->getArrayCopy());
+    }
+
+    public function testThatAttributeValuesWillBeEscaped(): void
+    {
+        $result = (string) ($this->helper)(['attribute' => '1&2']);
+        self::assertStringContainsString('attribute="1&amp;2"', $result);
+    }
+
+    public function testThatAttributeKeysWillBeEscaped(): void
+    {
+        $result = (string) ($this->helper)(['donkeys&goats' => 'value']);
+        self::assertStringContainsString('donkeys&amp;goats="value"', $result);
+    }
+}


Ocramius:
> the introduced HtmlAttributesSet class is untested and has several public methods

Was it introduced in `3.x`, or pre-existing?

In general, un-tested code is stopped at review time, so that may be an oversight.

froschdesign:
@gsteel 
> With the introduction of `$this->htmlAttributes()` a kind-of BC break was introduced…

That means we can never add a new view helper without to create a major version.

gsteel:
> @gsteel
> 
> > With the introduction of `$this->htmlAttributes()` a kind-of BC break was introduced…
> 
> That means we can never add a new view helper without to create a major version.

@froschdesign 

Not really. The mechanism for consumer code overriding standard helpers works fine. The problem is loading plugins from the helper manger by alias when you actually want to use a specific implementation. 

In this case, abstract helper assumes that the plugin with the alias `htmlAttributes` will not be overridden therefore it's safer to retrieve the plugin by is FQCN if you want to rely on its implemented functionality

froschdesign:
@gsteel 
We had the topic in the chat yesterday and the problem still exists.

In a perfect world, everyone would follow the documentation and register the helpers correctly, but unfortunately often only one version is registered under factories which means:

```php
$renderer = new Laminas\View\Renderer\PhpRenderer();
$renderer->getHelperPluginManager()->setService('htmlAttributes', static function() {
    return 'custom';
});

echo $renderer->htmlAttributes();
echo $renderer->htmlList(['foo', 'bar'], false, ['class' => 'baz']);
```

Output:

```php
custom

<ul class="baz">
<li>foo</li>
<li>bar</li>
</ul>
```

The HtmlList helper uses `Laminas\View\Helper\HtmlAttributes::class` and not my `htmlAttributes`.

The second problem I see is that we will be asked again why something is being done differently in this one place.

https://github.com/laminas/laminas-view/blob/7ba573a745c16b4359fac558b4ef8af7426943a8/src/Helper/AbstractHtmlElement.php#L67

gsteel:
Hi @froschdesign 

I have a slightly different opinion on all of this which doesn't matter, but my reasoning is this:

With the release of 2.13.0, the behaviour of `AbstractHtmlElement::htmlAttribs()` was changed from [this](https://github.com/laminas/laminas-view/blob/8d6e0961b7546d097d857dd529e33b545fde4463/src/Helper/AbstractHtmlElement.php#L65-L100) to [this](https://github.com/laminas/laminas-view/blob/2a5ce1b6dbaf2f965bad65852e309d6b7e8db4e2/src/Helper/AbstractHtmlElement.php#L65-L76). Where before 2.13 any helper that extended `AbstractHtmlElement` and made use of `htmlAttribs()` could assume that an attribute list would be correctly encoded. So prior to 2.13, calling `$renderer->htmlList(['foo', 'bar'], false, ['class' => 'baz']);` would yield the following whether you had a custom helper aliased to "htmlAttributes" or not.

```html
<ul class="baz">
<li>foo</li>
<li>bar</li>
</ul>
```

Because at the time of the 2.13 release I _did_ have a custom helper aliased to `htmlAttributes`, this change to `AbstractHtmlElement` caused segfaults in php.

Because, for a *very* long time, there has not been a helper for attributes present in the library, in many projects I have implemented one, because if you make stuff with html, it's convenient to call `$renderer->htmlAttributes($list)` in your views.

My reckoning is that I can't be the only person who does this.

The controversial change that I made in this pull, to call the helper by FQCN instead of its published short name restores behaviour prior to 2.13.

In my opinion, none of the helpers should be calling `$this->getView()->plugin('whatever')` and should all have constructors and factories where appropriate. For example, `HTMLList::__construct()` might look like

```php
public function __construct(Escaper $escaper, HtmlAttributes $attributeHelper) {}
```

On the other hand, if a plugin was consuming the HTMLList helper, then you might want to make sure that it used a 'custom' version of it, if one was available.

Anyhow, restoring behaviour of `AbstractHTMLElement::htmlAttribs()` seemed like the right thing to do to address the scenario I described in this pull, if it's not, then it can easily be reverted in a patch. For my sake, at least I know how to fix the segfaults that will occur in my projects when they update laminas-view.

froschdesign:
@gsteel 
> Because at the time of the 2.13 release I _did_ have a custom helper aliased to `htmlAttributes`, this change to `AbstractHtmlElement` caused segfaults in php.
> 
> Because, for a _very_ long time, there has not been a helper for attributes present in the library, in many projects I have implemented one, because if you make stuff with html, it's convenient to call `$renderer->htmlAttributes($list)` in your views.

Please keep in mind that a plugin system can always lead to these problems. And I'm on your side that a segfault is more than ugly but this only the symptom not the problem.

If you introduce a new plugin, then you either have to communicate this properly or you have to build something like a fallback or similar, _which unfortunately did not happen_.
And that is the real problem for me!

> The controversial change that I made in this pull, to call the helper by FQCN instead of its published short name restores behaviour prior to 2.13.

But it also leads to the problem I described. Yes, unfortunately the problem is _real_!
Many use the simplest and quickest way, and if it works, nothing is changed. So aliases are not registered.

> In my opinion, none of the helpers should be calling `$this->getView()->plugin('whatever')`

But it cannot be changed at this point.

froschdesign:
We need a command based on laminas-cli to create and register a view helper!

PR #89, opened 2021-10-13
Re-enable skipped tests and improve type coverage

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | maybe
| QA            | yes

### Description

Mostly just QA re-enabling tests and cleaning up modified tests for CS and psalm.

Potential BC break is changing ExceptionInterface to extend from Throwable - if this _is_ a BC break, I can revert and update baseline.

### DIFF
--- a/polyfill/Paginator/ArrayAdapter.php
+++ b/polyfill/Paginator/ArrayAdapter.php
@@ -14,6 +14,8 @@ use function count;
  * purposes of testing against PHP 8.1.
  *
  * @todo Remove when laminas-paginator has a release targeting PHP 8.1.
+ *
+ * @internal
  */
 class ArrayAdapter implements AdapterInterface
 {
--- a/polyfill/Paginator/ArrayAdapter.php
+++ b/polyfill/Paginator/ArrayAdapter.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Laminas\Paginator\Adapter;
+
+use ReturnTypeWillChange;
+
+use function array_slice;
+use function count;
+
+/**
+ * Test shim for PHP 8.1 compatibility
+ *
+ * This class replaces the laminas-paginator equivalent during development only, for
+ * purposes of testing against PHP 8.1.
+ *
+ * @todo Remove when laminas-paginator has a release targeting PHP 8.1.
+ */
+class ArrayAdapter implements AdapterInterface
+{
+    /**
+     * ArrayAdapter
+     *
+     * @var array
+     */
+    protected $array;
+
+    /**
+     * Item count
+     *
+     * @var int
+     */
+    protected $count;
+
+    /**
+     * @param array $array ArrayAdapter to paginate
+     */
+    public function __construct(array $array = [])
+    {
+        $this->array = $array;
+        $this->count = count($array);
+    }
+
+    /**
+     * Returns an array of items for a page.
+     *
+     * @param  int $offset Page offset
+     * @param  int $itemCountPerPage Number of items per page
+     * @return array
+     */
+    public function getItems($offset, $itemCountPerPage)
+    {
+        return array_slice($this->array, $offset, $itemCountPerPage);
+    }
+
+    /**
+     * Returns the total number of rows in the array.
+     *
+     * @return int
+     */
+    #[ReturnTypeWillChange]
+    public function count()
+    {
+        return $this->count;
+    }
+}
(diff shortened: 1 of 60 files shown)


gsteel:
Tests are failing on 8.1 because Paginator needs to be compatible with 8.1

froschdesign:
@gsteel 
> Tests are failing on 8.1 because Paginator needs to be compatible with 8.1

There are already some pollyfills:

https://github.com/laminas/laminas-view/tree/2.14.x/polyfill

Maybe the problem can be solved this way _for the moment_.

(I will create issue reports in the related component issue trackers to remove the pollyfills with an update.)

froschdesign:
@gsteel 
I found another problem:

https://github.com/laminas/laminas-view/blob/7ba573a745c16b4359fac558b4ef8af7426943a8/test/Helper/LayoutTest.php#L17-L34

If you could take a look at it, I would be very grateful.

gsteel:
> @gsteel
> 
> > Tests are failing on 8.1 because Paginator needs to be compatible with 8.1
> 
> There are already some pollyfills:
> 
> https://github.com/laminas/laminas-view/tree/2.14.x/polyfill
> 
> Maybe the problem can be solved this way _for the moment_.
> 
> (I will create issue reports in the related component issue trackers to remove the pollyfills with an update.)

Hi @froschdesign - I've added these polyfills and addressed the uninitialised props in Layout test

src/Exception/ExceptionInterface.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Laminas\View\Exception;
 
-interface ExceptionInterface
+use Throwable;
+
+interface ExceptionInterface extends Throwable
```
froschdesign, line 7: @Ocramius @boesing 
> Potential BC break is changing ExceptionInterface to extend from Throwable - if this is a BC break, I can revert and update baseline.

Any comments on this change?

froschdesign:
@gsteel 
> I've added these polyfills and addressed the uninitialised props in Layout test

I have already seen that. Great and thank you! 👍 

(I have the issue reports for the components on my list.)

src/Exception/ExceptionInterface.php
Ocramius: Potentially a BC break, but then `ExceptionInterface` is not supposed to be relied upon when implementing things downstream, but rather as a marker interface.

IMO acceptable, and I'm prepared to tell downstream consumers why they were wrong upfront, should there be problems :-)

polyfill/Paginator/ArrayAdapter.php
```diff
@@ -0,0 +1,65 @@
...
+use function array_slice;
+use function count;
+
+/**
+ * Test shim for PHP 8.1 compatibility
+ *
+ * This class replaces the laminas-paginator equivalent during development only, for
+ * purposes of testing against PHP 8.1.
+ *
+ * @todo Remove when laminas-paginator has a release targeting PHP 8.1.
```
Ocramius, line 16: Can we make this `@internal` here?

polyfill/Paginator/ArrayAdapter.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace Laminas\Paginator\Adapter;
```
Ocramius, line 3: We should `export-ignore` the `polyfill/` directory

Ocramius (changes requested):

Ocramius (approved):
LGTM :+1:

polyfill/Paginator/ArrayAdapter.php
froschdesign: @Ocramius 
But this means that in other environments laminas-view does not work with the paginator and PHP 8.1.
I am a bit surprised because it was not excluded before.

Ocramius:
Oooof, there's no `2.15.x` :| 

This sort of stuff should really be in a minor, but `laminas/automatic-releases` (correctly) didn't create `2.15.x`, because `3.0.x` already exists.

polyfill/Paginator/ArrayAdapter.php
Ocramius: @froschdesign aren't the `polyfill/` classes just for test use? They were only loaded in `autoload-dev`

froschdesign: You're absolutely right, of course!

Ocramius:
Thanks @gsteel!

PR #90, opened 2021-10-14
Null values in html attribute lists trigger unexpected exceptions

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| QA            | yes

Prior to 2.13.x, providing an array of attributes such as `['class' => 'whatever', 'data-something' => null]` to a helper that extends from `AbstractHtmlElement` would yield markup along the lines of `class="whatever" data-something=""`

Now, a type error is thrown in `HtmlAttributesSet::__toString()` causing a fatal error because of the the throw in __toString

### DIFF
--- a/src/Helper/HtmlAttributes.php
+++ b/src/Helper/HtmlAttributes.php
@@ -1,14 +1,15 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-view for the canonical source repository
- * @copyright https://github.com/laminas/laminas-view/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-view/blob/master/LICENSE.md New BSD License
- */
+declare(strict_types=1);
 
 namespace Laminas\View\Helper;
 
+use Laminas\Escaper\Escaper;
+use Laminas\View\Helper\Escaper\AbstractHelper as AbstractEscapeHelper;
 use Laminas\View\HtmlAttributesSet;
+use Laminas\View\Renderer\PhpRenderer;
+
+use function assert;
 
 /**
  * Helper for creating HtmlAttributesSet objects
@@ -18,12 +19,20 @@ class HtmlAttributes extends AbstractHelper
     /**
      * Returns a new HtmlAttributesSet object, optionally initializing it with
      * the provided value.
+     *
+     * @param iterable<string, scalar|array|null> $attributes
      */
     public function __invoke(iterable $attributes = []): HtmlAttributesSet
     {
+        $renderer = $this->getView();
+        assert($renderer instanceof PhpRenderer);
+        $escapePlugin = $renderer->plugin('escapeHtml');
+        assert($escapePlugin instanceof AbstractEscapeHelper);
+        $escaper = $escapePlugin->getEscaper();
+        assert($escaper instanceof Escaper);
+
         return new HtmlAttributesSet(
-            $this->getView()->plugin('escapehtml')->getEscaper(),
-            $this->getView()->plugin('escapehtmlattr')->getEscaper(),
+            $escaper,
             $attributes
         );
     }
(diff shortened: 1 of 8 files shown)


froschdesign:
> Prior to 2.13.x, providing an array of attributes such as `['class' => 'whatever', 'data-something' => null]` to a helper that extends from `AbstractHtmlElement` would yield markup along the lines of `class="whatever" data-something=""`

This was never tested either. :sob:

gsteel:
@froschdesign - I'm just working on fixing the HtmlAttributesSet class and providing tests for it. There will be BC breaks to HtmlAttributesSet, but considering it's been in the wild for less than a day, I think those breaks should be acceptable.

gsteel:
Hey @froschdesign, please could you review?

src/Helper/HtmlAttributes.php
```diff
@@ -18,12 +19,20 @@ class HtmlAttributes extends AbstractHelper
...
      */
     public function __invoke(iterable $attributes = []): HtmlAttributesSet
     {
+        $renderer = $this->getView();
+        assert($renderer instanceof PhpRenderer);
+        $escapePlugin = $renderer->plugin('escapeHtml');
+        assert($escapePlugin instanceof AbstractEscapeHelper);
+        $escaper = $escapePlugin->getEscaper();
+        assert($escaper instanceof Escaper);
+
```
gsteel, line 33: All those assertions keep psalm happy. Could we not just instantiate our own Escaper in __construct() and avoid all of that? i.e. `public function __construct(Escaper $escaper = null) { $this->escaper = $escaper ?? new Escaper(); }`

src/HtmlAttributesSet.php
```diff
@@ -21,40 +29,26 @@ class HtmlAttributesSet extends ArrayObject
...
-
-    /**
-     * Constructor.
-     *
-     * @param Escaper $htmlEscaper General HTML escaper
-     * @param Escaper $htmlAttributeEscaper Escaper for use with HTML attributes
-     * @param iterable $attributes Attributes to manage
-     */
-    public function __construct(Escaper $htmlEscaper, Escaper $htmlAttributeEscaper, iterable $attributes = [])
+    public function __construct(Escaper $escaper, iterable $attributes = [])
```
gsteel, line 34: How many escapers do we actually need?

src/Helper/HtmlAttributes.php
froschdesign: But this means that if the escaper is to be replaced, I have to add / write a factory for this helper too.
That would not be enough

```php
'factories' => [
    Helper\EscapeHtml::class     => MyEscaperFactory::class,
    Helper\EscapeHtmlAttr::class => MyEscaperFactory::class,
    Helper\EscapeJs::class       => MyEscaperFactory::class,
    Helper\EscapeCss::class      => MyEscaperFactory::class,
    Helper\EscapeUrl::class      => MyEscaperFactory::class,
]
```
(And that in itself is _unbeautiful_.)

src/HtmlAttributesSet.php
froschdesign: One, because there is only one class which offers all methods for HTML, HTML attributes, Javascript, CSS and URL.

https://docs.laminas.dev/laminas-escaper/intro/#overview

src/Helper/HtmlAttributes.php
gsteel: 👍 - I figured that others might balk at all the assertions.

src/HtmlAttributesSet.php
gsteel: Sorry, I was attempting a little bit of sarcasm there!

froschdesign: And I gave you an _expert answer_, so it worked! 😄

gsteel:
Good morning @froschdesign - Would it be possible to get this into a patch release today? I've got projects queued up for deploy that are failing CI due to this bug. Sorry to nag 😬

froschdesign (approved):

froschdesign:
@gsteel 
_Sorry for the delay. I have a lot of client projects at the moment._

Thank you for your time and this contribution! 👍

Issue #29, opened 2019-12-31
[v3] remove Zend\View\Stream and useStreamWrapper  option


weierophinney:
If I don't misunderstood, zf1 Zend_View_Stream was introduced for short_open_tag off environment , 

but `Since PHP 5.4.0, <?= is always available.`
http://php.net/manual/en/ini.core.php#ini.short-open-tag

so I think Zend_View_Stream can be removed since zf3.

---

Originally posted by @sasezaki at https://github.com/zendframework/zend-view/issues/23

weierophinney:
:+1: 


---

Originally posted by @gianarb at https://github.com/zendframework/zend-view/issues/23#issuecomment-164439614

PR #91, opened 2021-10-30
Declare `@deprecated` on the stream wrapper functionality (will be removed in `3.0.0`)

sasezaki:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

* refs. #29 
* also see https://github.com/zendframework/zend-view/pull/34
* This PR aims to provide deprecation for v2 
* I use `@deprecated` phpdoc in order to deprecation.  https://discourse.laminas.dev/t/deprecation-policy-trigger-error-vs-deprecated/2435

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -14,6 +14,8 @@ namespace Laminas\View;
  * written by
  *     Mike Naberezny (@link http://mikenaberezny.com)
  *     Paul M. Jones  (@link http://paul-m-jones.com)
+ *
+ * @deprecated will be removed in version 3
  */
 class Stream
 {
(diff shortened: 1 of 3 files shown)


Ocramius:
@froschdesign added you as reviewer, mostly to check if any docs need changing

src/Stream.php
```diff
@@ -14,6 +14,8 @@
  * written by
  *     Mike Naberezny (@link http://mikenaberezny.com)
  *     Paul M. Jones  (@link http://paul-m-jones.com)
+ *
+ * @deprecated will be removed in version 3
```
froschdesign:
```suggestion
 * @deprecated will be removed in version 3, only the normal PHP tags (<?php ?> and <?= ?>) should be used
```

sasezaki: @froschdesign Thanks suggestion. updated

Ocramius (approved):
Thanks @sasezaki!

PR #96, opened 2021-12-17
Replace `container-interop/container-interop` with `psr/container`

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | maybe
| QA            | yes

### Description

With reference to [this comment](https://github.com/mezzio/mezzio-laminasviewrenderer/pull/16#issuecomment-996010971) from @boesing, this pull removes all mention of Interop\Container

It doesn't appear that it is currently possible to have interop <1.2 installed due to SM >3.3 so the conflict in `composer.json` is largely irrelevant.

Because the lock file is updated, and newer version of psalm is installed, a number of non-trivial, new psalm errors emerged - I've added these initial errors to the baseline for now _(A lot of them are due to the deprecated and to be removed flash messenger)_.

WRT to BC, although types have changed from `Interop` to `Psr`, it wouldn't have been possible for any type to _not_ implement Psr prior to this patch would it?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "vimeo/psalm": "^4.10"
     },
     "conflict": {
+        "container-interop/container-interop": "<1.2",
         "laminas/laminas-servicemanager": "<3.3",
         "laminas/laminas-router": "<3.0.1",
         "zendframework/zend-view": "*"
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "laminas/laminas-paginator": "^2.5",
         "laminas/laminas-permissions-acl": "^2.6",
         "laminas/laminas-router": "^3.0.1",
-        "laminas/laminas-servicemanager": "^3.3",
+        "laminas/laminas-servicemanager": "^3.4",
         "laminas/laminas-session": "^2.12",
         "laminas/laminas-uri": "^2.5",
         "phpspec/prophecy": "^1.12",
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1329,9 +1329,8 @@
     </UndefinedInterfaceMethod>
   </file>
   <file src="src/Helper/Navigation/PluginManager.php">
-    <NonInvariantDocblockPropertyType occurrences="2">
+    <NonInvariantDocblockPropertyType occurrences="1">
       <code>$aliases</code>
-      <code>$factories</code>
     </NonInvariantDocblockPropertyType>
   </file>
   <file src="src/Helper/Navigation/Sitemap.php">
@@ -1949,9 +1948,8 @@
       <code>has</code>
       <code>has</code>
     </MixedMethodCall>
-    <NonInvariantDocblockPropertyType occurrences="2">
+    <NonInvariantDocblockPropertyType occurrences="1">
       <code>$aliases</code>
-      <code>$factories</code>
     </NonInvariantDocblockPropertyType>
     <UndefinedInterfaceMethod occurrences="2">
       <code>getServiceLocator</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
+<files psalm-version="v4.15.0@a1b5e489e6fcebe40cb804793d964e99fc347820">
   <file src="bin/templatemap_generator.php">
     <MissingParamType occurrences="1">
       <code>$templatePath</code>
@@ -656,8 +656,7 @@
       <code>$captureLock</code>
       <code>$captureType</code>
     </PropertyNotSetInConstructor>
-    <RedundantConditionGivenDocblockType occurrences="3">
-      <code>$this-&gt;view instanceof View\Renderer\RendererInterface</code>
+    <RedundantConditionGivenDocblockType occurrences="2">
       <code>(null !== $content) &amp;&amp; is_string($content)</code>
       <code>is_string($content)</code>
     </RedundantConditionGivenDocblockType>
@@ -706,24 +705,6 @@
     <PropertyNotSetInConstructor occurrences="1">
       <code>HeadTitle</code>
     </PropertyNotSetInConstructor>
-  </file>
-  <file src="src/Helper/HtmlAttributes.php">
-    <MixedArgument occurrences="2">
-      <code>$this-&gt;getView()-&gt;plugin('escapehtml')-&gt;getEscaper()</code>
-      <code>$this-&gt;getView()-&gt;plugin('escapehtmlattr')-&gt;getEscaper()</code>
-    </MixedArgument>
-    <MixedMethodCall occurrences="2">
-      <code>getEscaper</code>
-      <code>getEscaper</code>
-    </MixedMethodCall>
-    <PossiblyNullReference occurrences="2">
-      <code>plugin</code>
-      <code>plugin</code>
-    </PossiblyNullReference>
-    <UndefinedInterfaceMethod occurrences="2">
-      <code>plugin</code>
-      <code>plugin</code>
-    </UndefinedInterfaceMethod>
   </file>
   <file src="src/Helper/HtmlFlash.php">
     <MixedAssignment occurrences="1">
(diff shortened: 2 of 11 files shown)


src/Helper/Service/AssetFactory.php
```diff
@@ -2,17 +2,14 @@
...
 namespace Laminas\View\Helper\Service;
 
-use Interop\Container\ContainerInterface;
-use Laminas\ServiceManager\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
 use Laminas\View\Exception;
 use Laminas\View\Helper\Asset;
+use Psr\Container\ContainerInterface;
 
-class AssetFactory implements FactoryInterface
```
Ocramius, line 11: Looks like a BC break 🤔

Ocramius: Not sure about this one: practically, there shouldn't match depending on this, but this change makes it practically a BC break (like other removals of `implements`)

Ocramius (changes requested):

src/Helper/Service/AssetFactory.php
gsteel: I thought that too, but saw @boesing did a similar thing with the [Doctype factory](https://github.com/laminas/laminas-view/commit/d1d9b8ceda2d7845e03373eee9544806982c9531#diff-e6d323bcbed448c76f7d30b10217611dc0ffd9ab40ae27272dd9fe17d87da035) around 2.12 ish…

gsteel: I take it back. The factory was introduced in 1.12. The problem is that it prevents removing references to Interop if I revert.

boesing: No need to fully remove existing references. Only thing I wanted to ensure was that the docblock allows PSR-11 which will also include the interop version of the interface.

ofc we should try to get rid of most of the factory stuff to also get rid of the references to interop code but that wont be possible without a new major so fixing the docblock in the plugin manager should already provide a fix for the problem in the PR you mentioned in the initial post 👍🏻

src/Helper/Navigation/AbstractHelper.php
```diff
@@ -14,6 +13,7 @@
 use Laminas\View;
 use Laminas\View\Exception;
 use Laminas\View\Helper\TranslatorAwareTrait;
+use Psr\Container\ContainerInterface;
```
boesing: This will require us to bump this package to PHP 7.4 as we do change the argument type of `AbstractHelper#setServiceLocator` here.

This is only possible as of PHP 7.4 AFAIR, maybe @Ocramius has more insights here.

src/Helper/Navigation/PluginManager.php
```diff
@@ -15,15 +16,12 @@
...
-    /**
-     * @var string Valid instance types.
-     */
     protected $instanceOf = AbstractHelper::class;
 
     /**
      * Default aliases
      *
-     * @var string[]
+     * @inheritdoc
```
boesing: I personally prefer `array<string,string>` than `inheritdoc` since using inheritdoc always depends on the tooling which is used around:
- Some IDEs are not able to "merge" informations
- Codestandard allows "inheritdoc" even if there is nothing to inherit (e.g. when the parent property is removed)

So this could become less accurate in upstream projects when changing to `inheritdoc`. Not sure if we should do this.

src/Helper/Navigation/PluginManager.php
```diff
@@ -47,7 +45,7 @@ class PluginManager extends HelperPluginManager
     /**
      * Default factories
      *
-     * @var string[]
+     * @inheritdoc
```
boesing: I personally prefer `array<string,string>` than `inheritdoc` since using inheritdoc always depends on the tooling which is used around:
- Some IDEs are not able to "merge" informations
- Codestandard allows "inheritdoc" even if there is nothing to inherit (e.g. when the parent property is removed)

So this could become less accurate in upstream projects when changing to `inheritdoc`. Not sure if we should do this.

src/Helper/Service/AssetFactory.php
```diff
@@ -2,17 +2,14 @@
...
+class AssetFactory
```
boesing: We should re-add the factory here as removing existing factories might cause trouble in upstream projects.
Even tho, that I dont think factories should be extended (and thats why I always end up adding `final` to em), I've already seen factories which actually do extend other factories and thus changing anything here could be problematic.

I think removing the interface is fine until we keep the method signature.
Changing to PSR-11 container might be fine when bumping to PHP 7.4

boesing (changes requested):

src/Helper/Navigation/AbstractHelper.php
gsteel: I think that the bump to service manager ^3.4 solves that… Maybe…

src/Helper/Navigation/PluginManager.php
gsteel: I reverted, but I think the original reason I went with inheritdoc is because `array<string, string>` causes psalm errors - SM is annotated with `array<array-key, string>` in the version used, so I thought that inheritdoc would shortcut to the full config spec when View can catch up with more recent versions of Service Manager. Anyhow, I've changed to `@var array<string, class-string>` with a `psalm-var` of `<array-key, string> to keep psalm happy about the inherited inferred type. Hope that's the right course of action here.

composer.json
```diff
@@ -54,6 +54,7 @@
         "vimeo/psalm": "^4.10"
     },
     "conflict": {
+        "container-interop/container-interop": "<1.2",
```
Ocramius, line 57: Really like how this got solved, nice! :+1:

psalm-baseline.xml
```diff
@@ -3833,8 +3758,16 @@
     <UndefinedInterfaceMethod occurrences="1">
       <code>plugin</code>
     </UndefinedInterfaceMethod>
-    <UnevaluatedCode occurrences="1">
+    <UnevaluatedCode occurrences="11">
```
Ocramius, line 3737: Some of these `<UnevaluatedCode` issues appearing, but they don't seem to be problematic: baseline got reduced in size anyway

src/Helper/Navigation/PluginManager.php
```diff
@@ -15,15 +16,13 @@
...
-     * @var string Valid instance types.
-     */
     protected $instanceOf = AbstractHelper::class;
 
     /**
      * Default aliases
      *
-     * @var string[]
+     * @var array<string, class-string>
+     * @psalm-var array<array-key, string>
```
Ocramius: Keys must be `string` here - can't be `array-key`

Also unsure why the two types are misaligned for `@var` and `@psalm-var`: in both cases, this is `class-string`

src/HelperPluginManager.php
```diff
@@ -219,7 +220,7 @@ class HelperPluginManager extends AbstractPluginManager
      * helper works fine as an invokable. The factory for doctype simply checks for the
      * config value from the merged config.
      *
-     * @var array
+     * @var array<array-key, callable|string>
```
Ocramius: If this is inherited, I would suggest dropping the type declaration, for more precise type definitions to be inherited later on

Ocramius (approved):
LGTM

src/Helper/Navigation/PluginManager.php
gsteel: For the aliases, they are [annotated as string[]](https://github.com/laminas/laminas-servicemanager/blob/3.11.x/src/ServiceManager.php#L75) in SM. I removed @\inheritdoc as per [comment from](https://github.com/laminas/laminas-view/pull/96#discussion_r772277754) @boesing - I've added the correct type _(`array<string, string>`)_ and added the resulting error to the baseline. Does that work? Or would it be better to remove the docs and fix the type in SM?

src/HelperPluginManager.php
gsteel: Got it!

src/Helper/Navigation/PluginManager.php
Ocramius: @gsteel would be best to remove the docs here then, I suppose - the type in SM can be refined in a separate patch then

Ocramius (approved):
:ship: as-is

Further type refinements can be done later, but:

 * already reduces baseline
 * already upgrades to `psr/container`

I'd say the added value is here, and further discussions are mostly bike-shedding, and should be deferred to new patches.

Ocramius:
thanks @gsteel!

src/Helper/Navigation/PluginManager.php
Ocramius: Let's leave this as-is for now :)

PR #97, opened 2021-12-17
Deprecate `jsonExpressionFinder` usage in `Json` view helper

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | no
| New Feature   | kindof

### Description

As per [discussion](https://github.com/laminas/laminas-view/pull/68#issuecomment-789062431) about removing the dependency on `laminas-json` in #68 - This adds a deprecation warning when the relevant option is detected paving the way for removal in 3.0

### DIFF
--- a/docs/book/helpers/json.md
+++ b/docs/book/helpers/json.md
@@ -20,7 +20,21 @@ determine how to handle the content.
 <?= $this->json($this->data) ?>
 ```
 
-> ### Enabling encoding using Laminas\Json\Expr _(Deprecated)_
+> WARNING: **Deprecated**
+> 
+> ### Enabling encoding using Laminas\Json\Expr
+> 
+> **This feature of the Json view helper has been deprecated in version 2.16 and will be removed in version 3.0.**
+>
+> The JSON helper accepts an array of options that will be passed to `Laminas\Json\Json::encode()` and
+> used internally to encode data.
+> `Laminas\Json\Json::encode` allows the encoding of native JSON expressions using `Laminas\Json\Expr`
+> objects. This option is disabled by default. To enable this option, pass a boolean `true` to the
+> `enableJsonExprFinder` key of the options array:
+>
+> ```php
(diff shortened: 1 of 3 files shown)


docs/book/helpers/json.md
```diff
@@ -20,7 +20,7 @@ determine how to handle the content.
 <?= $this->json($this->data) ?>
 ```
 
-> ### Enabling encoding using Laminas\Json\Expr
+> ### Enabling encoding using Laminas\Json\Expr _(Deprecated)_
```
froschdesign:
```suggestion
> WARNING: **Deprecated**
> 
> ### Enabling encoding using Laminas\Json\Expr
> 
> **This feature of the Json view helper has been deprecated in version 2.16 and will be removed in version 3.0.**
>
> The JSON helper accepts an array of options that will be passed to `Laminas\Json\Json::encode()` and
> used internally to encode data.
> `Laminas\Json\Json::encode` allows the encoding of native JSON expressions using `Laminas\Json\Expr`
> objects. This option is disabled by default. To enable this option, pass a boolean `true` to the
> `enableJsonExprFinder` key of the options array:
>
> ```php
> <?= $this->json($this->data, ['enableJsonExprFinder' => true]) ?>
> ``
```

docs/book/helpers/json.md
```diff
@@ -31,3 +31,5 @@ determine how to handle the content.
 > ```php
 > <?= $this->json($this->data, ['enableJsonExprFinder' => true]) ?>
 > ```
+>
+> **This feature of the Json view helper has been deprecated in 2.16.x and will be removed in 3.x**
```
froschdesign:
```suggestion
```

Ocramius (approved):
Not a fan of `E_USER_DEPRECATED`, but in this case, it is true that we can't do it in a different way, due to the deprecation being bound to a specific value combination, rather than a specific type.

Ocramius:
Thanks @gsteel!

PR #99, opened 2021-12-30
Update to Laminas Coding Standard 2.3

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Updating and cleaning up for Laminas Coding Standard 2.3 including WIP on reducing Psalm Errors.

I appreciate that reviewing this is going to be hellish… FWIW, I'm pretty confident that there are no BC breaks but I have fixed some obvious errors in doc blocks.

The most notable changes are likely adding nullable type hints such as `function Foo(Type $baz = null)` changed to `function Foo(?Type $baz = null)`.

Also, there are a number of placeholder container/array access methods that had doc blocks for void return on methods such as `next()` that were returning `parent::next()` - as these methods should have always been returning void anyhow, the `return` statements have been dropped in these cases to satisfy CS.

Please let me know if you'd like me to squash.

I also got a bit cavalier with the psalm baseline - I was left with around 130 new errors that all looked like the result of changing line numbers rather than new errors, so I just updated/added to the baseline - ~1000 psalm errors have been fixed in this pull so I decided to quit whilst I was ahead…


Ocramius (changes requested):
Hmm, we'll need a `2.17.x` branch here, before merging

src/Helper/AbstractHtmlElement.php
```diff
@@ -1,22 +1,33 @@
...
      *
      * @deprecated just use PHP_EOL
      */
-    const EOL = PHP_EOL;
+    public const EOL = PHP_EOL;
 
     /**
      * The tag closing bracket
      *
      * @var string
```
Ocramius, line 28: Probably `string|null`?

src/Helper/HeadLink.php
```diff
@@ -488,11 +512,9 @@ public function createDataPrev(array $args)
      */
     public function createDataNext(array $args)
     {
-        $rel  = 'next';
-        $href = (string) array_shift($args);
-
-        $attributes = compact('rel', 'href');
```
Ocramius, line 494: Thanks for getting rid of `compact()`!

src/Helper/Navigation/Links.php
```diff
@@ -20,30 +45,30 @@ class Links extends AbstractHelper
...
+    public const RENDER_CUSTOM     = 0x8000;
+    public const RENDER_ALL        = 0xffff;
 
     /**
      * Maps render constants to W3C link types
      *
      * @var array
      */
-    protected static $RELATIONS = [
+    protected static $RELATIONS = [ // phpcs:ignore
```
Ocramius, line 71: Weird that PHPCS complains here

src/Helper/PaginationControl.php
```diff
@@ -20,29 +26,29 @@ class PaginationControl extends AbstractHelper
      *
      * @var string|array
      */
-    protected static $defaultViewPartial = null;
```
Ocramius, line 23: This had a `null` default, but the type doesn't match it (yes, I know the CS tooling just removed `= null`)

src/Renderer/ConsoleRenderer.php
```diff
@@ -85,7 +83,7 @@ public function setFilterChain(FilterChain $filters)
     public function getFilterChain()
     {
         if (null === $this->__filterChain) {
-            $this->setFilterChain(new FilterChain());
+            $this->__filterChain = new FilterChain();
```
Ocramius, line 86: Not much of a problem, but it may be possible that overrides to `setFilterChain()` make this change a potential issue.

I'll let this one roll though, as it's really an edge case.

src/Renderer/PhpRenderer.php
```diff
@@ -414,7 +429,7 @@ public function setFilterChain(FilterChain $filters)
     public function getFilterChain()
     {
         if (null === $this->__filterChain) {
-            $this->setFilterChain(new FilterChain());
+            $this->__filterChain = new FilterChain();
```
Ocramius, line 432: Similar to above location: setter may have been overridden, although I'll let this one through.

src/Resolver/PrefixPathStackResolver.php
```diff
@@ -46,6 +52,6 @@ public function resolve($name, Renderer $renderer = null)
             }
         }
 
-        return;
+        return; // phpcs:ignore
```
Ocramius, line 55: This really shouldn't be needed :thinking:

Ocramius (approved):
LGTM, thanks @gsteel!

src/Helper/AbstractHtmlElement.php
Ocramius: Not relevant for this patch - anyone willing to reduce the psalm baseline can touch this.

src/Helper/PaginationControl.php
Ocramius: Not relevant for this patch - anyone willing to reduce the psalm baseline can touch this.

src/Resolver/PrefixPathStackResolver.php
Ocramius: Non-blocker: still massive improvement, so keeping this here is no harm.

Ocramius:
Thanks @gsteel!

gsteel:
> Thanks @gsteel!

Thanks for reviewing @Ocramius! It was a bit of a monster that one 😅

Ocramius:
Yeah, but I was waiting while I was generating an almost 40Gb performance profile for an app: https://twitter.com/Ocramius/status/1479844147427172360 :D

PR #100, opened 2021-12-31
Use htmlattributes helper alias instead of hardcode class

neilime:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Currently AbstractHtmlElement uses hardcoded htmlattributes helper so we cannot use a custom helper by overding it via view_helpers config

This PR uses the helper alias

### DIFF
--- a/src/Helper/AbstractHtmlElement.php
+++ b/src/Helper/AbstractHtmlElement.php
@@ -64,7 +64,7 @@ abstract class AbstractHtmlElement extends AbstractHelper
             }
         }
 
-        $attribs = $this->getView()->plugin(HtmlAttributes::class)($attribs);
+        $attribs = $this->getView()->plugin('htmlattributes')($attribs);
 
         return (string) $attribs;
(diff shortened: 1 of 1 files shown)


src/Helper/AbstractHtmlElement.php
```diff
@@ -64,7 +64,7 @@ protected function htmlAttribs($attribs)
             }
         }
 
-        $attribs = $this->getView()->plugin(HtmlAttributes::class)($attribs);
+        $attribs = $this->getView()->plugin('htmlattributes')($attribs);
```
Ocramius, line 67: IMO this is not valid: if you wanted to have a custom plugin used, then the right approach would be to move thus to `__construct`, and then override how this specific helper is built.

Ocramius (changes requested):

PR #104, opened 2022-01-09
Cache and Session packages are not used anywhere

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Neither `cache` nor `session` are referenced anywhere in `view`.


Ocramius (approved):
Most excellent, thanks @gsteel!

Ocramius:
Hmm, lowest deps build is experiencing failures 🤔

Also, need to change default branch before merging (that's on me)

Ocramius (changes requested):
We'll need to handle this one: 


```
PHP Fatal error:  During inheritance of IteratorAggregate: Uncaught Return type of Laminas\Session\Storage\AbstractSessionArrayStorage::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```

No idea where it is pulling `laminas/laminas-session` from :|

gsteel:
I think lowest deps is failing here because flash messenger requires session ^2.8.x but ^2.12 is required for 8.1

Ocramius:
What we can do here then (possible alternatives):

 1. upgrade flash messenger to latest deps (separate patch) then upgrade flash messenger here
 2. introduce a `conflict` with older `laminas/laminas-session` versions

Ocramius (approved):
LGTM, thanks @gsteel!

PR #106, opened 2022-01-10
Update locked deps, refine types in `TemplatePathStack`

gsteel:
### Description

Updates locked versions of stdlib and flash messenger solving psalm errors introduced by type improvements therein. This will make further pulls to ^2 slightly less noisy.

### DIFF
--- a/src/Resolver/TemplatePathStack.php
+++ b/src/Resolver/TemplatePathStack.php
@@ -35,6 +35,8 @@ use const PATHINFO_EXTENSION;
 
 /**
  * Resolves view scripts based on a stack of paths
+ *
+ * @psalm-type PathStack = SplStack<string>
  */
 class TemplatePathStack implements ResolverInterface
 {
@@ -50,7 +52,7 @@ class TemplatePathStack implements ResolverInterface
      */
     protected $defaultSuffix = 'phtml';
 
-    /** @var SplStack */
+    /** @var PathStack */
     protected $paths;
 
     /**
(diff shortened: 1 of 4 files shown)


Ocramius (approved):
🚢  thanks @gsteel!

Ocramius:
Same story with the default branch to be fixed here 😬

src/Resolver/TemplatePathStack.php
```diff
@@ -236,13 +246,15 @@ public function addPath($path)
      */
     public function clearPaths()
     {
-        $this->paths = new SplStack();
+        /** @psalm-var PathStack $paths */
```
Ocramius, line 249: I don't understand what is going on here :thinking:

Ocramius (changes requested):

src/Resolver/TemplatePathStack.php
gsteel: I introduced a psalm-type on this class as `PathStack = SplStack<string>` - is that the wrong way to do it? The reason for the local variable here is so that I can tell psalm what it is before assigning to `$this->paths` otherwise I get a psalm error. Can you tell me a better way?

Ocramius: Ooooh, now I see it! It's all good, I was just confused about it, thanks for clarifying!

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #107, opened 2022-01-10
Remove Irrelevant MVC 2 Compat, since only `laminas/laminas-mvc:^3` is now installable

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

It's not possible to install view at any version more recent that ^2.16 in MVC versions before 3.x so this pull removes compat for versions prior to 3.x


Ocramius (approved):
Thanks @gsteel!

PR #108, opened 2022-01-10
Remove `laminas/laminas-servicemanager:^2` compatibility: only `^3` supported now

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

It's no longer possible for laminas-view to be used with service manager v2, so this pull cleans up code to provide compatibility with it.

### DIFF
--- a/src/Helper/Navigation/PluginManager.php
+++ b/src/Helper/Navigation/PluginManager.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Laminas\View\Helper\Navigation;
 
 use Interop\Container\ContainerInterface;
-use Laminas\ServiceManager\ConfigInterface;
 use Laminas\ServiceManager\Factory\InvokableFactory;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\View\HelperPluginManager;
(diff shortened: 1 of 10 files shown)


samsonasik (approved):

src/Helper/Navigation/PluginManager.php
```diff
@@ -67,20 +66,19 @@ class PluginManager extends HelperPluginManager
...
-     * @param array $v3config If $configOrContainerInstance is a container, this
-     *     value will be passed to the parent constructor.
+     * @param ContainerInterface $configOrContainerInstance
+     * @param array $v3config
      * @psalm-param ServiceManagerConfiguration $v3config
      */
     public function __construct($configOrContainerInstance = null, array $v3config = [])
     {
-        /** @psalm-suppress MissingClosureParamType */
+        /** @psalm-suppress UnusedClosureParam, MissingClosureParamType */
```
Ocramius, line 75: Note: one way to instruct psalm about unused parameters (which are still necessary) is to name them `$_something`, where `$_` indicates the intent of leaving them unused

Ocramius (approved):

src/Helper/Navigation/PluginManager.php
gsteel: Thanks @Ocramius, I didn't know that - The coding standard won't allow it though!

Ocramius (approved):
Thanks @gsteel!

PR #113, opened 2022-01-12
Declare required PHP extensions

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Declares `dom` and `filter` extensions as dependencies - surfaced by #112 - I'm guessing that #112 might not be merged unless we can allow failures? Otherwise I can introduce appropriate deps and ignore obvious aliases or those in `suggest` on that pull to make it pass

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,6 +1,7 @@
 {
     "extensions": [
         "dom",
+        "filter",
         "intl"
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
LGTM - good to have these declared, thanks @gsteel!

PR #111, opened 2022-01-12
Apply php 7.4 syntax + typed properties

samsonasik:
|    Q          |   A
|-------------- | ------
| QA            | yes

Apply php 7.4 syntax + typed properties

PR #114, opened 2022-01-12
Update `composer.lock` and remove `CHANGELOG.md`, to prevent issues with future merge-up PRs

gsteel:
I think that deps can't be installed on 2.x because of a merged pull without an updated lock file in 2.19 - should help with https://github.com/laminas/laminas-view/pull/111


Ocramius:
Thanks @gsteel!

PR #118, opened 2022-01-12
Workaround for Cyclic Incompatible Dependencies on 3.0.x

gsteel:
Adds a script to check out the current tree as a 2.99.x branch for CI runs so that composer will install cyclic incompatible dependencies.

### DIFF
--- a/.laminas-ci/pre-install.sh
+++ b/.laminas-ci/pre-install.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Temporary workaround for cyclic dependencies - can be removed once 3.0 has been released
+
+echo "Branch as 2.99.x in Pre-Install"
+
+git checkout -b 2.99.x


composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a003c15d0428178d28dd954a18766ec9",
+    "content-hash": "3052afdac31561cdc43064a657f79b64",
```
Ocramius: This will be an issue with merge-up operations: unsure if I should first merge up `2.20.0`, or merge this and then fix it at merge time.

gsteel: Ah yes, if we can get 2.20.x merged up, I can rebase and squash to get rid of all the noise

gsteel: …actually, I've squashed and removed the commit that changes the lock file. No checks will run because no PHP code has changed but once 2.20 is in, everything should be groovy. If it's easier, I can just merge 2.20 into this pull

Ocramius (approved):

Ocramius:
@gsteel the CI generated matrix looks really weird - is it bugged out?

gsteel:
@Ocramius - There's just zero checks because no *.php or *.md files have changed in this pull right? The checks ran before, but I squashed/forced pushed to clean up the pull - this is the last successful run before I squashed: https://github.com/laminas/laminas-view/actions/runs/1688557410

Ocramius:
Is it that? 🤔 

Really weird that it skips like that, but OK, I guess?

Ocramius:
Thanks @gsteel!

PR #126, opened 2022-01-31
Remove laminas-console dependency

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Closes #115 
WRT #101 

- Removes Console Model and Renderer
- Removes laminas-console from dev dependencies

PR #128, opened 2022-01-31
Remove legacy Zend compatibility and service names from plugin manager defaults

### LINKED ISSUE #117: Remove Zend Compatibility and Legacy Helper Aliases
### Feature Request

|    Q        |   A
|------------ | ------
| BC Break    | yes

#### Summary

All the zend related aliases are self explanatory, but the normalised aliases used back in ServiceManager 2.x such as `laminasviewhelperheadscript` should be removed for 3.x

I think it would also be worth settling on a consistent convention for the aliases so that instead of having `headScript`, `headscript` and `HeadScript` there is just one - I prefer `headScript`?

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Closes #117 

- Removes hard-coded, zend-related aliases from the various plugin managers
- Removes code that searches the container for Zend\Authentication in the Identity helper factory
- Cleans up the test for the identity helper factory to use regular PHP Unit mocks and also compatibility with ServiceManager 2.x which is long gone.
- Updates psalm baseline

_(Should also help to get CI to pass for #126)_


gsteel:
As this targets 3.0, I would also like to remove the v2 normalised helper aliases such as `laminasviewhelperdoctype` - Do that here or in a separate pull?

Ocramius:
> I would also like to remove the v2 normalised helper aliases such as `laminasviewhelperdoctype`

Unsure about these: would suggest looking at whether any docs reference them, first :thinking:

Ocramius (approved):
LGTM, thanks @gsteel!

PR #130, opened 2022-02-01
Chore: dot-file cleanup

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Changes docs-build workflow to avoid any future problems with permissions as per @weierophinney on Slack
- Removes coveralls config
- Updates ignores with changed/removed/added dot files

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -3,8 +3,7 @@ name: docs-build
 on:
   release:
     types: [published]
-  repository_dispatch:
-    types: docs-build
+  workflow_dispatch:
 
 jobs:
   build-deploy:
@@ -13,5 +12,5 @@ jobs:
       - name: Build Docs
         uses: laminas/documentation-theme/github-actions/docs@master
         env:
-          "DOCS_DEPLOY_KEY": ${{ secrets.DOCS_DEPLOY_KEY }}
-          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          DOCS_TOKEN: ${{ secrets.ORGANIZATION_ADMIN_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/docs-build.yml
```diff
@@ -3,8 +3,7 @@ name: docs-build
 on:
   release:
     types: [published]
-  repository_dispatch:
-    types: docs-build
+  workflow_dispatch:
```
Ocramius, line 6: Ref: https://laminas.slack.com/archives/CAP4J90LW/p1643662240275769

Quoting @weierophinney:

> Hi, all! We've noticed a few issues with the docs-build workflow recently. It doesn't appear to affect ALL repos, but when it does, you'll see one of the following symptoms:
> Action fails, and the logs indicate permission issues.
> In a few cases, I've tried to build and deploy locally, but those efforts generally result in the syntax highlighting going wonky (dark on dark... impossible to read!).
> I did some digging today, and figured out when and where the permissions issues occur, and a way to resolve them. It does *require changes* in the docs-build workflow; however, I've done the changes in a way that existing repos that do not have issues can continue to build as they have.
> For affected repositories, submit a PR with the following changes:
> In .github/workflows/docs-build.yml
> Find the line defining the env variable DOCS_DEPLOY_KEY
> And replace it with  DOCS_TOKEN: ${{ secrets.ORGANIZATION_ADMIN_KEY }}
> To allow maintainers/admins to test it out, you should also find the line repository_dispatch:
> And change it to workflow_dispatch:
> And then remove the line following (types: docs-build)
> If you do such a PR, drop a link in here so somebody with commit rights can merge and test.
> For those with merge rights, if you have the github CLI client, you can run gh workflow run docs-build.yml --ref {branch name where you pushed the changes to docs-build.yml}. It will build the most recent tag and push the results back to the gh_pages branch.
> As an example of one such repo, we'd had trouble with builds on mezzio-swoole, and the most recent build I did had resulted in bad syntax highlghting. Current docs were built with the workflow changes, and had zero issues.

Ocramius (approved):
Thanks @gsteel!

PR #139, opened 2022-02-13
Clean up `Escape*` helpers, deprecate extraneous methods that allow changing encoding/escaper on helpers

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| RFC           | yes
| QA            | yes

### Description

- The Escaper library is a hard dependency - moved from dev deps.
- The escaper helpers should only need to expose an __invoke method. Deprecate everything else.
- Provide transitional constructor that can be updated to _require_ a configured escaper provided by DI with encoding configured from config.view in 3.0
- Annotate helpers as final so they can be made final in 3.0
- Annotate abstract helper as internal

In 3.0 - I think that inheritance from the main abstract helper can be removed - there is no reason for these helpers to expose the renderer and they don't need it themselves.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,9 @@
         "sort-packages": true,
         "platform": {
             "php": "7.4"
+        },
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true
         }
     },
     "require": {
@@ -26,6 +29,7 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-json": "*",
+        "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-json": "^3.3",
         "laminas/laminas-stdlib": "^3.6"
@@ -34,7 +38,6 @@
         "laminas/laminas-authentication": "^2.5",
         "laminas/laminas-coding-standard": "~2.3.0",
         "laminas/laminas-console": "^2.6",
-        "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-feed": "^2.15",
         "laminas/laminas-filter": "^2.13.0",
         "laminas/laminas-http": "^2.15",
--- a/src/Helper/Escaper/AbstractHelper.php
+++ b/src/Helper/Escaper/AbstractHelper.php
@@ -32,13 +32,13 @@ abstract class AbstractHelper extends Helper\AbstractHelper
     /** @var Escape|null */
     protected $escaper;
 
-    /** @psalm-suppress DeprecatedProperty */
+    /**
+     * @param non-empty-string $encoding
+     * @psalm-suppress DeprecatedProperty
+     */
     public function __construct(?Escape $escaper = null, string $encoding = 'UTF-8')
     {
-        if ($escaper) {
-            $this->escaper = $escaper;
-        }
-
+        $this->escaper  = $escaper;
         $this->encoding = $escaper ? $escaper->getEncoding() : $encoding;
     }
 
@@ -102,7 +102,7 @@ abstract class AbstractHelper extends Helper\AbstractHelper
      *
      * @deprecated Will be removed in 3.0.
      *
-     * @param  string $encoding
+     * @param  non-empty-string $encoding
      * @throws Exception\InvalidArgumentException
      * @return AbstractHelper
      * @psalm-suppress DeprecatedProperty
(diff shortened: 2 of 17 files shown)


composer.json
```diff
@@ -34,7 +38,6 @@
         "laminas/laminas-authentication": "^2.5",
         "laminas/laminas-coding-standard": "~2.3.0",
         "laminas/laminas-console": "^2.6",
-        "laminas/laminas-escaper": "^2.5",
```
Ocramius, line 37: Oh wow, this was no good 😱

src/Helper/Escaper/AbstractHelper.php
```diff
@@ -14,23 +14,34 @@
...
+    protected $encoding;
 
-    /** @var Escaper\Escaper|null */
+    /** @var Escape|null */
     protected $escaper;
 
+    /** @psalm-suppress DeprecatedProperty */
+    public function __construct(?Escape $escaper = null, string $encoding = 'UTF-8')
+    {
+        if ($escaper) {
```
Ocramius: This conditional is not necessary

src/Helper/Escaper/AbstractHelper.php
```diff
@@ -14,23 +14,34 @@
...
+     * @var string
+     */
+    protected $encoding;
 
-    /** @var Escaper\Escaper|null */
+    /** @var Escape|null */
     protected $escaper;
 
+    /** @psalm-suppress DeprecatedProperty */
+    public function __construct(?Escape $escaper = null, string $encoding = 'UTF-8')
```
Ocramius, line 39: We should probably require the encoding to be a `non-empty-string`

Ocramius (changes requested):

composer.json
```diff
@@ -19,13 +19,17 @@
...
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true
         }
     },
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-json": "*",
+        "laminas/laminas-escaper": "^2.5",
```
Ocramius, line 32: This is a change big enough that we should release 2.20 asap, IMO 🤔

Ocramius (approved):
Thanks @gsteel!

src/Helper/Escaper/AbstractHelper.php
```diff
@@ -89,9 +100,12 @@ abstract protected function escape($value);
     /**
      * Set the encoding to use for escape operations
      *
-     * @param  string $encoding
+     * @deprecated Will be removed in 3.0.
```
froschdesign, line 103: @gsteel @Ocramius 
This is a no-go, because as long as these methods are present, it can be used and overwritten. Therefore, we must assume that this will be done.
This means when a deprecation is introduced that we also have to explain what the alternative is. The current description is therefore insufficient. _It is not user-friendly!_
(It would be perfect if we could add a link to the documentation with further explanations and an example.)

Ocramius: The idea is to drop the abstract class overall? 🤔

gsteel: @froschdesign Won't the migration guide cover this?

Should the migration guide to 3 be seeded in 2.x so that the docs can be written and referenced for each deprecation?

gsteel: > The idea is to drop the abstract class overall?

I was actually referring to `Helper\AbstractHelper` in the pull description. `Helper\Escaper\AbstractHelper` IMO could be concrete and final so that the user can choose:

```php
// Single Helper for all escaping contexts
$this->escaper()->escapeJs('foo');
$this->escaper()->escapeHtml('foo');

// Existing helpers for convenience, or maybe just get rid of them?
$this->escapeHtml('foo'); // etc...
```

The useful stuff in `Helper\Escaper\AbstractHelper` with regard to recursion could be extracted into a trait for re-use.

froschdesign: > Won't the migration guide cover this?

So far this is not the case and if the migration guide covers this, the link should be added here.

I see two options for deprecations like this:

1. add an explanation of what the alternative is / what the user should do instead in the DocBlock
2. create a migration guide like "Migration from Version 2.x to 2.20" and add the link to the DocBlock

The goal must be that the user does not have to search or ask. The past has shown that questions arise in the forum, in the chat and new issue reports if we do not clarify things properly.
Thank you for understanding!

(We must not forget the release notes too, the explanations and / or the link must also be added there.)

PR #140, opened 2022-02-14
Clean up Asset helper and factory deprecating extraneous methods

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Introduces a trait that will be helpful to finding and removing inheritance from AbstractHelper
- Fixes possibly null return value of `getView()` in `HelperInterface`
- Deprecates runtime set/get of the Asset helper's resource map
- Improves the exception message for undefined assets
- Marks both the helper and the factory as @\final in doc block
- Slight change to behaviour of the factory so that an exception is only thrown if the resource map is configured as anything other than an array. Previously, an exception would be thrown for empty configuration
- General improvements to psalm type inference

### DIFF
--- a/src/Helper/Asset.php
+++ b/src/Helper/Asset.php
@@ -7,31 +7,49 @@ namespace Laminas\View\Helper;
 use Laminas\View\Exception;
 
 use function array_key_exists;
+use function sprintf;
 
 /**
  * View helper plugin to fetch asset from resource map.
+ *
+ * @final
  */
 class Asset extends AbstractHelper
 {
-    /** @var array */
+    use DeprecatedAbstractHelperHierarchyTrait;
+
+    /** @var array<non-empty-string, non-empty-string> */
     protected $resourceMap = [];
 
     /**
-     * @param string $asset
-     * @return string
+     * @param array<non-empty-string, non-empty-string> $resourceMap
+     */
+    public function __construct(array $resourceMap = [])
+    {
+        $this->resourceMap = $resourceMap;
+    }
+
+    /**
+     * @param non-empty-string $asset
+     * @return non-empty-string
      * @throws Exception\InvalidArgumentException
      */
     public function __invoke($asset)
     {
         if (! array_key_exists($asset, $this->resourceMap)) {
-            throw new Exception\InvalidArgumentException('Asset is not defined.');
+            throw new Exception\InvalidArgumentException(sprintf(
+                'The asset with the name "%s" has not been defined.',
+                $asset
+            ));
         }
 
         return $this->resourceMap[$asset];
     }
 
     /**
-     * @param array $resourceMap
+     * @deprecated The Resource map should be provided to the constructor from version 3.0
+     *
+     * @param array<non-empty-string, non-empty-string> $resourceMap
      * @return $this
      */
     public function setResourceMap(array $resourceMap)
@@ -42,7 +60,9 @@ class Asset extends AbstractHelper
     }
 
     /**
-     * @return array
+     * @deprecated
+     *
+     * @return array<non-empty-string, non-empty-string>
      */
     public function getResourceMap()
     {
--- a/src/Helper/Asset.php
+++ b/src/Helper/Asset.php
@@ -47,7 +47,8 @@ class Asset extends AbstractHelper
     }
 
     /**
-     * @deprecated The Resource map should be provided to the constructor from version 3.0
+     * @deprecated since 2.20.0, this method will be removed in version 3.0.0 of this component.
+     *             The Resource map should be provided to the constructor from version 3.0
      *
      * @param array<non-empty-string, non-empty-string> $resourceMap
      * @return $this
@@ -60,7 +61,8 @@ class Asset extends AbstractHelper
     }
 
     /**
-     * @deprecated
+     * @deprecated since 2.20.0, this method will be removed in version 3.0.0 of this component.
+     *             Runtime retrieval of the resource map from the view will be removed without replacement.
      *
      * @return array<non-empty-string, non-empty-string>
      */
--- a/src/Helper/HelperInterface.php
+++ b/src/Helper/HelperInterface.php
@@ -18,7 +18,7 @@ interface HelperInterface
     /**
      * Get the View object
      *
-     * @return Renderer
+     * @return Renderer|null
      */
     public function getView();
 }
--- a/src/Helper/Service/AssetFactory.php
+++ b/src/Helper/Service/AssetFactory.php
@@ -9,8 +9,12 @@ use Laminas\ServiceManager\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
 use Laminas\View\Exception;
 use Laminas\View\Helper\Asset;
+use Traversable;
 
+use function gettype;
 use function is_array;
+use function iterator_to_array;
+use function sprintf;
 
 /**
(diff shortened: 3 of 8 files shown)


src/Helper/AbstractHelper.php
```diff
@@ -18,7 +18,7 @@ abstract class AbstractHelper implements HelperInterface
     /**
      * Set the View object
      *
-     * @return $this
+     * @return self
```
gsteel, line 21: Psalm is happier with this

src/Helper/Asset.php
```diff
@@ -7,31 +7,49 @@
...
      * @throws Exception\InvalidArgumentException
      */
     public function __invoke($asset)
     {
         if (! array_key_exists($asset, $this->resourceMap)) {
-            throw new Exception\InvalidArgumentException('Asset is not defined.');
+            throw new Exception\InvalidArgumentException(sprintf(
+                'The asset with the name "%s" has not been defined.',
+                $asset
+            ));
```
gsteel, line 43: Is this a potential security issue? No attempt is made to escape the key, so if the user is showing exception messages and allowing untrusted input for the asset key, this could cause issues, or is it 'not our responsibility' if the user chooses to show exception messages in production?

src/Helper/HelperInterface.php
```diff
@@ -18,7 +18,7 @@ public function setView(Renderer $view);
     /**
      * Get the View object
      *
-     * @return Renderer
+     * @return Renderer|null
```
gsteel, line 21: All implementors tend to extend from AbstractHelper where the return type is plausibly null

src/Helper/Service/AssetFactory.php
```diff
@@ -22,24 +26,23 @@ class AssetFactory implements FactoryInterface
...
         $config = $container->get('config');
-        if (isset($config['view_helper_config']['asset'])) {
-            $configHelper = $config['view_helper_config']['asset'];
-            if (isset($configHelper['resource_map']) && is_array($configHelper['resource_map'])) {
-                $helper->setResourceMap($configHelper['resource_map']);
-            } else {
-                throw new Exception\RuntimeException('Invalid resource map configuration.');
-            }
-        }
+        $config = is_array($config) ? $config : [];
```
gsteel, line 37: Should this include `instanceof ArrayAccess` then `iterator_to_array`

test/Helper/Service/AssetFactoryTest.php
```diff
@@ -53,22 +59,78 @@ public function testValidConfiguration(): void
...
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider validConfigProvider
+     * @param array<string, mixed> $config
+     */
+    public function testThatAnExceptionWillNotBeThrownWhenGivenUnsetOrEmptyArrayConfiguration(array $config): void
```
gsteel, line 129: Replaces `testValidConfiguration()` which was deprecated

src/Helper/AbstractHelper.php
Ocramius: Seems to work fine? :thinking: 

https://psalm.dev/r/29584fecc0

src/Helper/Asset.php
Ocramius: > so if the user is showing exception messages

This would be the security issue: showing infrastructural exception messages is never supposed to happen, so `Asset.php` is not responsible for such a fault.

src/Helper/AbstractHelper.php
gsteel: It's to avoid generating more errors in other implementors because of the trait:
https://psalm.dev/r/331a66aa21

src/Helper/Asset.php
gsteel: Thanks. I thought so… but it was worth making sure!

src/Helper/AbstractHelper.php
Ocramius: Makes sense.

src/Helper/Service/AssetFactory.php
```diff
@@ -12,6 +12,10 @@
 
 use function is_array;
 
+/**
+ * @final
+ * @psalm-suppress DeprecatedInterface Compatibility with Service Manager 2 should be removed in version 3.0.
```
froschdesign, line 21: I think the dependency to laminas-servicemanager can also be removed in version 3.

The same applies to all other factories:

https://github.com/laminas/laminas-view/blob/5e8d8002811234e99dbaf021ab62007b5cfc22b3/src/Helper/Service/FlashMessengerFactory.php#L12
_(The flash messenger helper will be removed anyway.)_

https://github.com/laminas/laminas-view/blob/5e8d8002811234e99dbaf021ab62007b5cfc22b3/src/Helper/Service/IdentityFactory.php#L14

See the factory for the Doctype helper:

https://github.com/laminas/laminas-view/blob/3855f6ef2c7f5c514ba0c0e9a1876e4f42ec8384/src/Helper/Service/DoctypeFactory.php#L7-L12

gsteel: Absolutely - and we can also say bye-bye to Interop too.

gsteel:
@Ocramius – Maybe this one can go into 2.20x too if there's time? The trait for deprecating extension of `AbstractHelper` will be useful in development of 3.0 for killing off pointless inheritance.

src/Helper/Service/AssetFactory.php
```diff
@@ -22,24 +26,23 @@ class AssetFactory implements FactoryInterface
...
+        $resourceMap = $this->assertArray('resource_map', $helperConfig);
 
-        return $helper;
+        return new Asset($resourceMap);
     }
 
     /**
      * Create service
      *
+     * @deprecated
```
Ocramius: We should explain the deprecation, in line with other deprecation comment requirements that @froschdesign made clear :)

src/Helper/Service/AssetFactory.php
Ocramius: Not sure about this one: it is very possible that `$config` is an `ArrayAccess`, so IMO we should `assert(is_array($config) || $config instanceof ArrayAccess)` here :thinking:

src/Helper/Service/AssetFactory.php
```diff
@@ -48,4 +51,18 @@ public function createService(ServiceLocatorInterface $serviceLocator, $rName =
...
+
+    /**
+     * @param array<array-key, mixed> $array
+     * @return array<array-key, mixed>
+     */
+    private function assertArray(string $key, array $array): array
+    {
+        $value = $array[$key] ?? [];
+        if (! is_array($value)) {
+            throw new Exception\RuntimeException('Invalid resource map configuration.');
```
Ocramius: This exception doesn't tell us which key is missing and/or malformed - this exception will just be confusing

Ocramius (changes requested):

src/Helper/Service/AssetFactory.php
gsteel: I've gone for Traversable here which I'm pretty sure is inline with most other components, because that allows a straight-forward iterator_to_array. OK?

froschdesign: > …is inline with most other components…

Correct.

Ocramius (approved):
Thanks @gsteel!

PR #141, opened 2022-02-14
Improve Deprecation Messages for Escape Helpers

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| RFC           | yes
| QA            | yes

### Description

@froschdesign - Do the following alterations to the doc blocks and the addition of factories mitigate your concerns WRT #139 

- [x] Unit test new factories
- [ ] <del>Implement Factories for other Escape* helpers</del>
- [ ] <del>Consider different encodings per helper context or not as the case may be…</del>


src/Helper/Service/EscaperFactoryTrait.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /** @return non-empty-string */
+    private function viewEncoding(ContainerInterface $container): string
+    {
+        $config = $container->has('config')
+            ? $container->get('config')
+            : [];
+
+        assert(is_array($config));
+        $viewConfig = $this->assertArray('view_manager', $config);
```
gsteel: Is `view_manager` traditionally where the `encoding` value is to be found? I can't remember… I think it may have been dropped entirely because I cannot find it in the MVC skeleton anymore…

src/HelperPluginManager.php
```diff
@@ -256,7 +256,7 @@ class HelperPluginManager extends AbstractPluginManager
         Helper\EscapeHtml::class          => InvokableFactory::class,
         Helper\EscapeHtmlAttr::class      => InvokableFactory::class,
         Helper\EscapeJs::class            => InvokableFactory::class,
-        Helper\EscapeCss::class           => InvokableFactory::class,
+        Helper\EscapeCss::class           => Helper\Service\EscapeCssFactory::class,
```
gsteel: All of the helpers are still new-able without a factory. Should factory addition be delayed for the inclusion of a Config Provider in #121 or for v3?

src/Helper/Escaper/AbstractHelper.php
```diff
@@ -124,7 +127,7 @@ public function setEncoding($encoding)
     /**
      * Get the encoding to use for escape operations
      *
-     * @deprecated Will be removed in 3.0.
+     * @deprecated Encoding is now considered a configuration item and not relevant in a template context
```
froschdesign: The information of the version was good, so it should be kept.

Suggestion:

```suggestion
     * @deprecated since 2.20.0, this method will be removed in version 3.0.0 of this component.
                   Encoding is now considered a configuration item and not relevant in a template context
```

src/Helper/Escaper/AbstractHelper.php
```diff
@@ -137,7 +140,9 @@ public function getEncoding()
     /**
      * Set instance of Escape
      *
-     * @deprecated The escaper should be provided to the constructor. This method will be removed in 3.0.
+     * The escaper is now provided to the constructor by factories for each specific helper type.
+     *
+     * @deprecated This method will be removed in 3.0.
```
froschdesign:
```suggestion
     *
     * @deprecated since 2.20.0, this method will be removed in 3.0.0.
                   The escaper is now provided to the constructor by factories for each specific helper type.
```

gsteel:
@froschdesign - With the more helpful annotations, is implementing the factories a requirement in your opinion or not? I'd prefer to skip the factories for now and add them in 3.0 when these methods are removed. However, if you want the alternatives implemented for 2.20.x, I'll write the factories and tests now.

src/Helper/Service/EscaperFactoryTrait.php
froschdesign: You are right, there is no reference to it anywhere. But it is a _hidden feature_, so someone may use it.

froschdesign:
@gsteel 
>  I'd prefer to skip the factories for now and add them in 3.0 when these methods are removed.

Then let's do it that way! 👍

src/Helper/Service/EscaperFactoryTrait.php
gsteel: Thanks for confirming. This is exactly what should end up in a ConfigProvider for this component such as #121

gsteel:
@froschdesign - Thank you for the review. I've removed the concrete factory but retained the currently unused trait, marked it as internal and added unit tests for it - that way, the work is preserved for use in 3.0

Ocramius (changes requested):
@gsteel my main concern with this patch is that the `trait` is introduced, but never used.

While I can see that it is `@internal`, both as a maintainer and a consumer of the library I do not understand its purpose, since we don't touch it.

gsteel:
@Ocramius – It's purpose was for use in factories to be added… in future… It's gone now!

froschdesign (approved):

Ocramius (approved):
Thanks for cleaning up then :D

PR #142, opened 2022-02-19
Refactor `Doctype`

gsteel:
Instead of introducing a V3 helper here _(Why did I do that?)_, I've re-applied the changes to the existing Doctype helper.

- Removes inheritance
- Adds types
- Adds the `final` keyword
- Removes all runtime mutation
- Changes default doctype to HTML5
- Removes the ability to use custom doctypes
- Removed methods
  - `setView`
  - `getView`
  - `setDoctype`
  - `getDoctype`
  - `getDoctypes`
  - `unsetDoctypeRegistry`
- implements `Stringable`

A number of tests rely on runtime mutation of the current doctype - these have been patched in a pretty ugly way for the time-being.

All of these helpers will eventually be refactored to use constructor injection of the Doctype helper so they can modify their output based on the configured doctype.

Examples are `HeadScript`, `HeadStyle` etc.


src/Helper/DoctypeV3.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        return $this;
+    }
+
+    public function __toString(): string
+    {
+        return $this->doctypeDeclaration();
+    }
+
+    /** @return non-empty-string */
```
Ocramius: IMO `key-of<self::$docTypeDefinitions>`

src/Helper/DoctypeV3.php
```diff
@@ -0,0 +1,112 @@
...
+    public const XHTML1_RDFA11       = 'XHTML1_RDFA11';
+    public const XHTML_BASIC1        = 'XHTML_BASIC1';
+    public const XHTML5              = 'XHTML5';
+    public const HTML4_STRICT        = 'HTML4_STRICT';
+    public const HTML4_LOOSE         = 'HTML4_LOOSE';
+    public const HTML4_FRAMESET      = 'HTML4_FRAMESET';
+    public const HTML5               = 'HTML5';
+
+    /** @var array<string, non-empty-string> */
+    private static $docTypeDefinitions = [
```
Ocramius: Can be a constant

src/Helper/DoctypeV3.php
```diff
@@ -0,0 +1,112 @@
...
+            ));
+        }
+
+        return self::$docTypeDefinitions[$id];
+    }
+
+    /**
+     * Is doctype XHTML?
+     */
+    public function isXhtml(?string $doctypeId = null): bool
```
Ocramius: Also here, let's restrict these input types as much as possible

Ocramius (changes requested):

src/Helper/DoctypeV3.php
gsteel: Nice - Didn't know that!

src/Helper/Doctype.php
```diff
@@ -32,199 +31,95 @@ class Doctype extends AbstractHelper
...
-
-        return $doctypes[$this->getDoctype()];
+        return $this->doctypeDeclaration();
     }
 
     /**
-     * Register the default doctypes we understand
-     *
-     * @return void
+     * @param DoctypeID|string|null $doctypeId
```
Ocramius, line 78: Thr `|string` kinda kills all type safety introduced here though

Ocramius (approved):

src/Helper/Doctype.php
gsteel: Yes - this was mainly so that the defensive check/exception was worth doing. I figured the check was worth retaining because of less precise inputs from a view script context.

Theoretically, no-one will interact with that method anyhow, unless they are using the helper as a lookup table, and it'll crash for invalid input.

gsteel: Am hoping that I can ship a Psalm plugin and collector (or whatever they're called) so we can get SA on .phtml files BTW

docs/book/v3/helpers/doctype.md
```diff
@@ -18,38 +17,31 @@ The `Doctype` helper allows you to specify one of the following types:
...
-easiest approach is to specify it in your bootstrap:
+The `Doctype` helper requires a constant indicating the desired doctype to its constructor.
+It can then be cast to a string in order to emit the doctype declaration:
 
 ```php
 use Laminas\View\Helper\Doctype;
 
-$doctypeHelper = new Doctype();
-$doctypeHelper->doctype('XHTML1_STRICT');
+$helper = new Doctype(Doctype::HTML5);
```
froschdesign: HTML5 is the default doctype therefore it should be omitted.
A second example with a different doctype would be good.

docs/book/v3/helpers/doctype.md
```diff
@@ -79,13 +71,14 @@ return [
 ];
 ```
 
+!NOTE The default doctype is HTML 5 when no configuration is specified.
```
froschdesign: The syntax is wrong but I would explain this with a second example, see above.

> ## Basic Usage
> 
> Start the admonition with one of the types in capital letters followed by a
> colon.
> The type corresponds with a specific style.
> The type should be followed by the content of the admonition.
> 
> ```markdown
> NOTE: An example of a note.
> ```
>
> ## Type of Admonitions
> 
> | Title / Topic             | Type    |
> |---------------------------|---------|
> | New Feature               | TIP     |
> | Deprecation Warning       | CAUTION |
> | Installation Requirements | MISSING |
> | General Notes             | NOTE    |
> | Security Warning          | WARNING |

froschdesign (changes requested):

froschdesign (approved):

PR #145, opened 2022-02-21
Added `laminas/laminas-servicemanager` as non-dev dependency, required at runtime

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| QA            | yes

### Description

`Psr\Container` and `Interop\Container` are too…

This is something I thought would be good to get into 2.20 before it gets released - In 3.0, if we introduce a config provider, this will mean that we'll be able to import the psalm types for factory configuration etc. I think that `^3.6` is the feasible minimum for service menager for this, currently though, I'd find it unlikely that installing view on PHP 7.4 gets you anything less that `3.10` of service manager.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,10 +29,13 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-json": "*",
+        "container-interop/container-interop": "^1.2",
         "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-json": "^3.3",
-        "laminas/laminas-stdlib": "^3.6"
+        "laminas/laminas-servicemanager": "^3.10",
+        "laminas/laminas-stdlib": "^3.6",
+        "psr/container": "^1 || ^2"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -29,10 +29,13 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-json": "*",
+        "container-interop/container-interop": "^1.2",
```
Ocramius, line 32: Do we need a dependency on `container-interop`, or could perhaps just keep `psr/container`?

Ocramius (changes requested):

composer.json
gsteel: It's imported by code in `src/` but I'll drop it if you'd like me to?

Ocramius: Is it possible to surgically remove it without breaking BC? :thinking:

gsteel: I don't think so, because it's using factory interfaces from service manager that declare it as an argument :( - AFAICT, it can only be dropped when we kill the factories in 3.

Ocramius (approved):
Thanks @gsteel!

PR #146, opened 2022-02-21
Helper and Inheritance Deprecations

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes
| QA            | yes

### Description

- Json Helper: Deprecate inheritance tree, deprecate `setResponse()` so it's effectively just `json_encode()` _(Eventually)_ - The response really shouldn't be modified in the view should it?
- HtmlAttributes Helper: Deprecate inheritance tree - it only needs an escaper to do its job
- The HTML List helper - also only needs access to an escaper
- <del>The Gravatar helper is deprecated completely with all its setters and getters. An alternative implementation is provided that simply returns an  tag based on its arguments</del> Moved to #147
- The view model helper doesn't need its current inheritance tree
- The layout helper only needs access to the view model helper so it is changed to pave the way for factory based constructor injection of the dependent helper

There's no need IMO to have this block a 2.20 release

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -19,7 +19,7 @@
     </MixedMethodCall>
   </file>
   <file src="src/Helper/AbstractHtmlElement.php">
-    <MissingConstructor occurrences="13">
+    <MissingConstructor occurrences="12">
       <code>$closingBracket</code>
       <code>$closingBracket</code>
       <code>$closingBracket</code>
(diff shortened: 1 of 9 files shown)


Ocramius:
Let's split this up: gravatar added in smells a bit 😁

Ocramius (approved):
LGTM, thanks @gsteel!

Few notes of things that may be improved in future:

 * when checking against `Foo|null`, instead of `! $foo`, `$foo === null` is a better approach, as it will be picked up by static analysis, if the type changes
 * when casting `object` to `string`, use `$object->__toString()` rather than `(string) $object`, which avoids having regressions when `$object` (by accident) is no longer an `object`

PR #148, opened 2022-02-22
Refactor identity helper

gsteel:
### Description

The identity helper imports laminas auth which makes it an undeclared runtime dependency.

Because only this helper (and its factory) uses the laminas auth package, the helper has been altered to check for the methods `hasIdentity` and `getIdentity`.

Submitting the pull without test alterations in the first commit so that CI can pass before the tests are changed


Xerkus:
Laminas Authentication is a soft dependency. Interface does not have to be declared to be used for checks as in the existing code. No autoload attempt is happening when instanceof check is used.

Removing interface reduces type safety for no gain.

src/Helper/Identity.php
```diff
@@ -4,57 +4,70 @@
...
-use Laminas\View\Exception;
+use function method_exists;
 
 /**
  * View helper plugin to fetch the authenticated identity.
  */
 class Identity extends AbstractHelper
 {
-    /**
-     * AuthenticationService instance
```
Ocramius, line 16: IMO to be reverted: this code does not initialize an `AuthenticationSetvice`, but accepts an already existing one.

A type level declaration is preferable to going back to `object`, even if it leads to a soft dependency.

Stubbing it out could suffice, but the removal of it from `composer.json` potentially leads to conflicts with new major releases of the authentication component. Perhaps replacing `require` with `conflict` would suffice here.

Ocramius (changes requested):

gsteel:
@Xerkus and @Ocramius - Reverted the duck typing so it's just a general clean up and deprecation of runtime get/set of the auth service.

src/Helper/Service/IdentityFactory.php
```diff
@@ -41,23 +38,30 @@ public function createService(ServiceLocatorInterface $serviceLocator, $rName =
...
-     */
-    private function discoverAuthenticationService(ContainerInterface $container)
+    private function discoverAuthenticationService(ContainerInterface $container): ?AuthenticationServiceInterface
     {
-        if ($container->has(AuthenticationService::class)) {
-            return $container->get(AuthenticationService::class);
-        }
+        // phpcs:disable WebimpressCodingStandard.Formatting.StringClassReference
+        $search = [
+            AuthenticationService::class,
```
Ocramius, line 45: IMO start from the `*Interface`, then fallback to concrete classes :thinking:

Ocramius: Hmm, I see this was the previous behavior (checking concrete implementation first) - nvm :|

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @gsteel!

gsteel:
Thanks too to you @Ocramius for your help and reviews of all this stuff 😃

PR #153, opened 2022-03-23
Fix #7 - placeholder view helper return value should be given also with `null` input

### LINKED ISSUE #7: Fix Placeholder view helper return value
This PR aims to fix the bug, outlined in the following issue: https://github.com/zendframework/zend-view/issues/16

The changes will allow `$this` as a return value of the view helper, when no name is passed as an argument, thus providing access to the methods inside the view helper. For this reason, more tests have been added to `ZendTest\View\Helper\PlaceholderTest`.

---

Originally posted by @thexpand at https://github.com/zendframework/zend-view/pull/162

### LINKED ISSUE #32: Placeholder View Helper "containerExists" makes no sense
At the moment the public function containerExists does not have any change to work:

because __invoke checks if a name is set it is impossible to call placeholder()
So to check if a placeholder name exists you have to all call 
placeholder('foobar')->containerExists('foobar')

but then the container is created in __invoke and containerExists returns in every case true.

And because placeholder('dummy') returns a container containerExists is never reachable and usable.

---

Originally posted by @bernhard-efler at https://github.com/zendframework/zend-view/issues/16

gsteel:
Rebase/Re-Create #49 

Fixes #7 
Fixes #32 
Closes #49

### DIFF
--- a/src/Helper/Placeholder.php
+++ b/src/Helper/Placeholder.php
@@ -37,7 +37,10 @@ class Placeholder extends AbstractHelper
      *
      * @param  string $name
      * @throws InvalidArgumentException
-     * @return AbstractContainer
+     * @return AbstractContainer|self
+     * @psalm-template T of string|null
+     * @psalm-param T $name
+     * @psalm-return (T is null ? self : AbstractContainer)
      */
     public function __invoke($name = null)
     {
--- a/src/Helper/Placeholder.php
+++ b/src/Helper/Placeholder.php
@@ -42,9 +42,7 @@ class Placeholder extends AbstractHelper
     public function __invoke($name = null)
     {
         if ($name === null) {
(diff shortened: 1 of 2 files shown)


src/Helper/Placeholder.php
```diff
@@ -37,14 +37,15 @@ class Placeholder extends AbstractHelper
      *
      * @param  string $name
      * @throws InvalidArgumentException
-     * @return AbstractContainer
+     * @return AbstractContainer|self
+     * @psalm-template T of string|null
+     * @psalm-param T $name
+     * @psalm-return (T is null ? self : AbstractContainer)
```
Ocramius, line 43: Nice! :)

Ocramius (approved):
LGTM 👍

PR #155, opened 2022-03-23
Resolve deprecations in Identity view helper

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

Removes deprecated methods and inheritance in the Identity helper, marking it final.
Removes factory interface inheritance and swaps service manager for container interface
Fixes factory tests to test helper behaviour instead of calling removed service getter.


Ocramius (approved):
Thanks @gsteel!

PR #147, opened 2022-02-22
Deprecates and replaces the Gravatar helper with a simplified version…

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### Description

The current Gravatar view helper has a tonne of setters and getters, dodgy inheritance and checks out superglobals to figure out if the current http request is secure or not.

The existing helper is deprecated completely in this pull with a replacement implementation that simply spits out an `img` tag with the relevant `src`.

Notable differences…
- Uses protocol relative url for the `src` to avoid inspecting the request
- Closes the img tag with /> regardless of whether the current doc is XHTML/HTML5

Because this was originally part of #146 - it imports the psalm type `AttributeSet` from HtmlAttributesSet which is added in #146 so psalm qa will fail.

Maybe this pull can be left for when/if 146 is merged and then rebased so psalm passes.

PR #156, opened 2022-03-23
Use DI for the HtmlAttributes helper

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no/maybe
| New Feature   | yes
| QA            | yes

### Description

The HTML Attributes helper was reaching into the view (which might not be there) in order to fetch a plugin (Escape helper), so that it could fetch an escaper instance from that plugin… and it was messy.

This pull adds a factory that injects an escaper into the helpers constructor and wires it up in the plugin manager.

This will aid in dealing with future changes to the shipped escape helpers.

It also deprecates the inheritance in the helper itself.

Not sure if this is considered BC breaking or not, theoretically, consumers will only interact with the helper by way of `$this->htmlAttributes()` and that works the same as before.

I could make the escaper nullable in the constructor and new one if preferred and then tighten things up further in 3 when we deal with the deprecations?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
 /.gitattributes export-ignore
 /.github/ export-ignore
 /.gitignore export-ignore
+/.laminas-ci export-ignore
 /.laminas-ci.json export-ignore
 /docs/ export-ignore
 /mkdocs.yml export-ignore
--- a/.laminas-ci/pre-install.sh
+++ b/.laminas-ci/pre-install.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Temporary workaround for cyclic dependencies - can be removed once 3.0 has been released
+
+echo "Branch as 2.99.x in Pre-Install"
+
+git checkout -b 2.99.x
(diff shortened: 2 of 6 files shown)


gsteel:
Not sure why composer is refusing to install. This pull was branched from 2.21.x and is targeting it too 🤷‍♂️

Are we going to have to back-port the `pre-install.sh` to checkout 2.99.x before install?

Ocramius:
I think this is because the branch is on a PR. The checkout may be needed until circular dependencies are gone :|

gsteel:
Could we get #146 in `2.21.x` before looking at this one? This conflicts with #146 and I'd rather rebase this on that than the other way around!

Ocramius:
@gsteel I had asked to split that one up - do you prefer to have it in there in one go?

gsteel:
@Ocramius I already did split it up - the gravatar stuff went into #147

gsteel:
So changes to the helper in #146 were nearly identical to this pull, so now, this is only the factory addition - also the constructor argument is reverted to be nullable, meaning it can be instantiated without the factory at all, therefore preserving BC

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #157, opened 2022-03-24
Cleanup BasePath View Helper

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes
| QA            | yes

### Description

- Mark helper as final and deprecate inherited methods / inheritance
- Add constructor that accepts the configured base path
- Deprecate runtime modification of the base path
- Improve helper tests and cover uncovered paths
- Introduce and wire up a factory for the helper that retrieves the base path from the expected config as per MVC docs

### DIFF
--- a/src/Helper/BasePath.php
+++ b/src/Helper/BasePath.php
@@ -4,22 +4,28 @@ declare(strict_types=1);
 
 namespace Laminas\View\Helper;
 
-use Laminas\View\Exception;
+use Laminas\View\Exception\RuntimeException;
 
 use function ltrim;
 use function rtrim;
+use function sprintf;
 
 /**
  * Helper for retrieving the base path.
+ *
+ * @final
  */
 class BasePath extends AbstractHelper
 {
-    /**
-     * Base path
-     *
-     * @var string
-     */
+    use DeprecatedAbstractHelperHierarchyTrait;
+
+    /** @var string|null */
     protected $basePath;
+
+    public function __construct(?string $basePath = null)
+    {
+        $this->basePath = $basePath;
+    }
 
     /**
      * Returns site's base path, or file with base path prepended.
@@ -27,24 +33,31 @@ class BasePath extends AbstractHelper
      * $file is appended to the base path for simplicity.
      *
      * @param  string|null $file
-     * @throws Exception\RuntimeException
+     * @throws RuntimeException
      * @return string
      */
     public function __invoke($file = null)
     {
-        if (null === $this->basePath) {
-            throw new Exception\RuntimeException('No base path provided');
+        if ($this->basePath === null) {
+            throw new RuntimeException('No base path provided');
         }
 
-        if (null !== $file) {
-            $file = '/' . ltrim($file, '/');
+        if (! empty($file)) {
+            return sprintf(
+                '%s/%s',
+                rtrim($this->basePath, '/'),
+                ltrim($file, '/')
+            );
         }
 
-        return $this->basePath . $file;
+        return rtrim($this->basePath, '/');
     }
 
     /**
      * Set the base path.
+     *
+     * @deprecated since 2.21.0, this method will be removed in version 3.0.0 of this component.
+     *             The base path should be provided to the constructor from version 3.0
      *
      * @param  string $basePath
      * @return self
--- a/src/Helper/BasePath.php
+++ b/src/Helper/BasePath.php
@@ -42,7 +42,7 @@ class BasePath extends AbstractHelper
             throw new RuntimeException('No base path provided');
         }
 
-        if (! empty($file)) {
+        if ($file !== null && $file !== '') {
             return sprintf(
                 '%s/%s',
                 rtrim($this->basePath, '/'),
--- a/src/Helper/BasePath.php
+++ b/src/Helper/BasePath.php
@@ -56,9 +56,6 @@ class BasePath extends AbstractHelper
     /**
      * Set the base path.
      *
-     * @deprecated since 2.21.0, this method will be removed in version 3.0.0 of this component.
-     *             The base path should be provided to the constructor from version 3.0
-     *
      * @param  string $basePath
      * @return self
      */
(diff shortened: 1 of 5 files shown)


src/Helper/BasePath.php
```diff
@@ -4,48 +4,61 @@
...
     {
-        if (null === $this->basePath) {
-            throw new Exception\RuntimeException('No base path provided');
+        if ($this->basePath === null) {
+            throw new RuntimeException('No base path provided');
         }
 
-        if (null !== $file) {
-            $file = '/' . ltrim($file, '/');
+        if (! empty($file)) {
```
Ocramius: `empty()` should generally never be used 😬

src/Helper/BasePath.php
```diff
@@ -4,48 +4,61 @@
...
         }
 
-        return $this->basePath . $file;
+        return rtrim($this->basePath, '/');
     }
 
     /**
      * Set the base path.
      *
+     * @deprecated since 2.21.0, this method will be removed in version 3.0.0 of this component.
```
Ocramius: One design problem here: how does one run a long-running MVC application with this basePath fixed? 🤔

Ocramius (changes requested):

src/Helper/BasePath.php
gsteel: I did consider the long running application scenario, because it's a target for all the work I'm doing. I personally think that if you're in a position where the removal of runtime changes to the base path affect you, then you've probably already written a custom helper to do this job. There's no existing mechanism to change the base path on a per-request basis. In MVC and Mezzio, it's the kind of thing that's done once and the presence of the method at all smells like the lack of DI from back in the ZF1 days.

Happy to revert, but I personally think that changing it during runtime is a routing concern, whereas the purpose of this helper is, as I understand it, to provide a stable prefix to _all_ routes or files. ??

src/Helper/BasePath.php
gsteel: 😂 I thought you'd hate that! I don't really see the problem with it. I know it's a string, or it's null, and I just want to check for a non-empty-string - I reckon it's more readable than the alternative `if ($file !== null && $file !== '') {`… `(string) $file !== ''` doesn't work here for psalm.

src/Helper/BasePath.php
froschdesign: > I personally think that if you're in a position where the removal of runtime changes to the base path affect you, then you've probably already written a custom helper to do this job.

The goal cannot be for everyone to write their own helper when long-running applications are used.
So the question is if we can solve the problem here at all.

gsteel: I believe that the presence and purpose of the `setWhatever()` method here is about favouring setter injection in older versions of this component. My argument is that _if_ you are using View in a long running app, you are already in a whole world of hurt and you would have to already be dealing with all the state problems it brings. Moving it to the constructor and prohibiting runtime modification at least nails up the intentions of the original design (As I understand it) so that we can in future work out how view can be used successfully in long running apps. Maybe a clear definition of what base path is for would be helpful. Base path for static files? For everything? In its current state, I can't see how it would be a useful tool for multiple, different base paths, bearing in mind you get the same instance every time you fetch it from the container (By default)

Ocramius: In this case, it's really not for setter injection, but because of the stateful nature of this helper.

gsteel: I investigated this in MVC and it looks like the setting is indeed only used at construction time in a factory [link](https://github.com/laminas/laminas-mvc/blob/ae0dbea9aed39dd6cb85f998b5ca4df320df90fd/src/Service/ViewHelperManagerFactory.php#L110)… but I've reverted the deprecation anyway as removing it doesn't really provide much benefit.

src/Helper/Service/BasePathFactory.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $config = $container->has('config')
+            ? $container->get('config')
+            : [];
+
+        assert(is_array($config) || $config instanceof ArrayAccess);
+
+        // The expected location in config for the base path in an MVC application
+        // is config.view_manager.base_path
+        // @link https://docs.laminas.dev/laminas-mvc/services/#viewmanager
```
froschdesign, line 27: And in a Mezzio application?

src/Helper/BasePath.php
```diff
@@ -4,43 +4,53 @@
...
+        if ($file !== null && $file !== '') {
+            return sprintf(
+                '%s/%s',
+                rtrim($this->basePath, '/'),
+                ltrim($file, '/')
+            );
         }
 
-        return $this->basePath . $file;
+        return rtrim($this->basePath, '/');
```
froschdesign: Looks like `rtrim` is used three times with the same options. `setBasePath` can be also used in the constructor.

src/Helper/Service/BasePathFactory.php
gsteel: I've been thinking about this, Mezzio and MVC have disparate config keys for the same things, particularly templates. the `view_manager` top level key, `view_helper_config`, `view_helpers` all have cross over I think. In an ideal world view would define a single top level key of `laminas_view` with everything underneath - maybe we can sort this out for 3.0 and then leave the Mezzio/MVC integration packages to sort out their own config structure. Anyhow, for the purposes of this pull, from searching around, I found `view_manager.base_path` to be the conventional default. Not convinced that base path is even useable by default in Mezzio as per recent Slacks…

froschdesign: > …maybe we can sort this out for 3.0…

laminas-mvc-view is the way we should go.

src/Helper/BasePath.php
gsteel: Sorry… That was sloppy!

froschdesign: No problem at all! 👍

Ocramius (approved):
LGTM! I will merge+release this after https://github.com/laminas/automatic-releases/pull/210 is through.

Thanks @gsteel!

PR #158, opened 2022-03-24
Add missing documentation for the replacement Gravatar Image Helper

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

Further to #147 - Adds docs for the replacement Gravatar helper along with deprecation notices for the replaced helper.

### DIFF
--- a/docs/book/helpers/gravatar-image.md
+++ b/docs/book/helpers/gravatar-image.md
@@ -1,0 +1,46 @@
+# GravatarImage
+
+The `GravatarImage` helper is useful for rendering image HTML markup returned from
+the [gravatar.com](https://gravatar.com) service.
+
+## Basic Usage
+
+You can use the `GravatarImage` helper anywhere in view scripts per the following example:
+
+```php
+echo $this->gravatarImage('email@example.com');
+```
+
+The first argument passed to the helper should be an e-mail address for which you want grab an avatar from gravatar.com. For convenience, this e-mail will be automatically hashed via the md5 algorithm.
+
+This will render an HTML img tag similar to the following:
+
+```html
+<img src="//www.gravatar.com/avatar/5658ffccee7f0ebfda2b226238b1eb6e?s=80&d=mm&r=g" />
+```
+
+## Custom Settings
+
+You can customize the request and HTML output for a gravatar.com image by passing additional arguments to the view helper:
+
+```php
+
+// Basic usage provides a 80px square image, falling back to the "mm" default image provided by the Gravatar service
+$this->gravatarImage('email@example.com');
+
+// Set the image size you want gravatar.com to return, in pixels
+$this->gravatarImage('email@example.com', 120);
+
+// Provide custom HTML attributes for the generated img tag:
+$this->gravatarImage('email@example.com', 120, [
+    'alt' => 'Profile Picture for Someone',
+    'data-something' => 'other-thing',
+]);
+
+// Set the default avatar image to use if gravatar.com does not find a match
+$this->gravatarImage('email@example.com', 120, [], 'identicon');
+
+// Set the avatar "rating" threshold (often used to omit NSFW avatars)
+$this->gravatarImage('email@example.com', 120, [], 'mm', 'g');
+
+```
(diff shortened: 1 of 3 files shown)


docs/book/helpers/gravatar-image.md
```diff
@@ -0,0 +1,46 @@
...
+    'data-something' => 'other-thing',
+]);
+
+// Set the default avatar image to use if gravatar.com does not find a match
+$this->gravatarImage('email@example.com', 120, [], 'identicon');
+
+// Set the avatar "rating" threshold (often used to omit NSFW avatars)
+$this->gravatarImage('email@example.com', 120, [], 'mm', 'g');
+
+```
```
froschdesign: Code blocks with longer descriptions must be omitted. It helps not the user, the search engine and navigation.
Please create separated sections with headline, explanation and code example.

Example:

> ## Set Image Size
> 
> The basic usage provides a 80px square image, falling back to the "mm" default image provided by the Gravatar service.
> To set a custom image size, use the second parameter and specify the size in pixels:
> 
> ```php
> $this->gravatarImage('email@example.com', 120);
> ```
> 
> Output:
> 
> ```html
> 
> ```
> 
> ## Set Custom HTML Attributes
> 
> …

_The helper has public constants for the image values and ratings, which seem not to be usable here, unfortunately. Or I'm wrong here?_

gsteel: The constants are no longer there for individual strings, only lists of possible valid values. They are typed as `value-of<self::CONSTANT>` so psalm will catch unlisted values.

froschdesign: _Sorry for the very late response, I was busy._

> …only lists of possible valid values.

Then a user must know all allowed values. So this means the _helper_ does not _help_ me here. 🤔  

Until now I did not know a single option:

- 404: do not load any image if none is associated with the email hash, instead return an HTTP 404 (File Not Found) response
- mp: (mystery-person) a simple, cartoon-style silhouetted outline of a person (does not vary by email hash)
- identicon: a geometric pattern based on an email hash
- monsterid: a generated 'monster' with different colors, faces, etc
- wavatar: generated faces with differing features and backgrounds
- retro: awesome generated, 8-bit arcade-style pixelated faces
- robohash: a generated robot with different colors, faces, etc
- blank: a transparent PNG image

gsteel: Hey @froschdesign, I'll re-implement the individual constants in another pull request and then update the docs here

docs/book/helpers/gravatar-image.md
```diff
@@ -0,0 +1,80 @@
...
+<img src="..." alt="Profile Picture for Someone" data-something="other-thing" />
+```
+
+### Change the fallback image
+
+The Gravatar service will present a default image when a given email address does not correspond to a known profile picture. The possible values are listed in `GravatarImage::DEFAULT_IMAGE_VALUES` and [documented here](https://en.gravatar.com/site/implement/images/). Provide this string value as the 4th argument.
+
+```php
+// Set the default avatar image to use if gravatar.com does not find a match
+echo $this->gravatarImage('email@example.com', 120, [], 'identicon');
```
froschdesign: The [_new_ constants](https://github.com/laminas/laminas-view/pull/162) can be used here.

docs/book/helpers/gravatar-image.md
```diff
@@ -0,0 +1,80 @@
...
+echo $this->gravatarImage('email@example.com', 120, [], 'https://example.com/default-image.png');
+```
+
+### Change the image rating allowed
+
+The Gravatar service allows users to provide a rating for the images they upload to indicate the type of audience they should be acceptable to. By default, the rating is "G". You can allow potentially explicit profile images by changing the rating to a value as [documented by the Gravatar service](https://en.gravatar.com/site/implement/images/), providing it as the 5th argument:
+
+```php
+// Set the avatar "rating" threshold (often used to omit NSFW avatars)
+$this->gravatarImage('email@example.com', 120, [], 'mp', 'pg');
```
froschdesign: Also here.

Ocramius (approved):
:ship:

PR #159, opened 2022-03-24
Server URL Helper Cleanup

gsteel:
### Description

Deprecates all methods in the server url helper and introduces a factory to provide the helper with the detected or configured host URI.

Brings in a hard dependency on laminas-diactoros in order to re-use the functions for detecting the host/port from the environment that are already present there.

Doesn't break BC according to tests, and also doesn't alter existing behaviour for long-running apps because the previous implementation caches the host/port in instance properties anyway.

The main improvement here is that consumers can configure the desired host and port without any attempt to figure it out from the environment which is particularly useful on the CLI or any other time when the required env vars are either wrong or not there at all.

The introduction of the factory is a potential BC break because an exception is thrown when there is neither configuration, nor usable env vars.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "ext-filter": "*",
         "ext-json": "*",
         "container-interop/container-interop": "^1.2",
+        "laminas/laminas-diactoros": "^2.8",
         "laminas/laminas-escaper": "^2.5",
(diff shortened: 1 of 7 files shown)


gsteel:
Closing - This is a duplicate of #160 - I got all messed up with GHA being down

PR #160, opened 2022-03-24
Server URL Helper Cleanup

gsteel:
### Description

Deprecates all methods in the server url helper and introduces a factory to provide the helper with the detected or configured host URI.

Brings in a hard dependency on laminas-diactoros in order to re-use the functions for detecting the host/port from the environment that are already present there.

Doesn't break BC according to tests, and also doesn't alter existing behaviour for long-running apps because the previous implementation caches the host/port in instance properties anyway.

The main improvement here is that consumers can configure the desired host and port without any attempt to figure it out from the environment which is particularly useful on the CLI or any other time when the required env vars are either wrong or not there at all.

The introduction of the factory is a potential BC break because an exception is thrown when there is neither configuration, nor usable env vars.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "ext-filter": "*",
         "ext-json": "*",
         "container-interop/container-interop": "^1.2",
+        "laminas/laminas-diactoros": "^2.8",
         "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-json": "^3.3",
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "ext-filter": "*",
         "ext-json": "*",
         "container-interop/container-interop": "^1.2",
-        "laminas/laminas-diactoros": "^2.8",
+        "laminas/laminas-diactoros": "^2.13.0",
         "laminas/laminas-escaper": "^2.5",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-json": "^3.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,107 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f9740d25a5b9d0482c3dedf4adc31831",
+    "content-hash": "7b5b413ef38742957ad478a7d0f2a92a",
     "packages": [
+        {
+            "name": "laminas/laminas-diactoros",
+            "version": "2.13.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/laminas/laminas-diactoros.git",
+                "reference": "34ba65010be9aa74e159d168c5ecfa5c01e4d956"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/laminas/laminas-diactoros/zipball/34ba65010be9aa74e159d168c5ecfa5c01e4d956",
+                "reference": "34ba65010be9aa74e159d168c5ecfa5c01e4d956",
+                "shasum": ""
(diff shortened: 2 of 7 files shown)


gsteel:
Oh… GHA is down again…

Ocramius:
@gsteel I re-opened this: was it closed on purpose?

gsteel:
@Ocramius - GHA confused me by being down and I had a duplicate pull - I left this one open and closed the other one.

Ocramius:
@gsteel if you can rebase this, I can throw it into `2.21.0`

composer.json
```diff
@@ -30,6 +30,7 @@
         "ext-filter": "*",
         "ext-json": "*",
         "container-interop/container-interop": "^1.2",
+        "laminas/laminas-diactoros": "^2.8",
```
Ocramius: Ah, wait... this landed in `require` now? :thinking:

Ocramius:
Nvm, needs some work, removing milestone for now, and releasing.

composer.json
```diff
@@ -30,6 +30,7 @@
         "ext-filter": "*",
         "ext-json": "*",
         "container-interop/container-interop": "^1.2",
+        "laminas/laminas-diactoros": "^2.13.0",
```
Ocramius, line 33: I don't think we should drag in a dependency for this :thinking:

Ocramius (changes requested):

composer.json
gsteel: I'm with you there, but I considered it better than attempting to parse `$_SERVER` inside this particular view helper.

Perhaps just abandon the patch and remove the helper instead - i.e. make sure that both MVC and Mezzio have their own implementations to replace it prior to a release of 3.0.

There are probably other helpers that could do with the same treatment - thinking `url` and perhaps `baseUrl` at the very least.

Ocramius: Overall, I think this does bring value, and is implemented the right/safe way, but it also leads to the added dependency, which is a lot :thinking: 

Not sure what to do here. /cc @weierophinney maybe?

weierophinney: I would likely copy over that functionality here and refactor to work with laminas-http. Since Diactoros is a PSR-7 library, there's a mismatch with the primary target of laminas-view (laminas-mvc). There's a reason we provide alternate url() and serverUrl() helpers in the mezzio-template-laminasview package...

gsteel: Can we/I create an entirely new library to contain everything that's MVC specific so that it can be stripped out in 3.0 - this could include all the supporting code around MVC events. If that's an agreeable path, I could just abandon this patch and deprecate the helper here entirely instead?

weierophinney: I think that would be perfect, @gsteel!

froschdesign: And we have to make sure that there is also a solution to create URLs with the same helpers when using laminas-cli. The server URL and the base path must be able to set via configuration, factory or something else.

weierophinney: @froschdesign I'd argue that should be via separate helpers. Not quite sure how to manage it (right now, CLI config is the same as web config), but having separate helpers for separate contexts makes a lot of sense here to me.

gsteel: @weierophinney 

I've made a start on this with just the ServerUrl and Url helpers specifically for MVC:

https://github.com/gsteel/laminas-mvc-view/pull/1

If this is a good enough starting point, is transferring the repo ownership an option?

weierophinney: Yes - we usually vote on any new packages in the TSC; this sort of thing can happen between meetings, though. Feel free to initiate; contact me in Slack for tips on how to do it.

Looking forward to further decoupling!

gsteel:
Closing… With https://github.com/laminas/laminas-mvc-view/pull/1 in progress, I suspect we'll simply delete the Url and ServerUrl helpers here in 3.0 release.

PR #163, opened 2022-04-06
Introduces an Empty ConfigProvider and Module

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes

### Description

Adds a `Module` and `ConfigProvider` that are effectively empty.

It's not possible to move all of the view helper configuration into the config provider without breaking BC.

Having them at least present and setup in composer.json for the component installer means that consumers will get used to having the provider/module injected into app config, thereby providing a way to introduce factories for various things on the journey to version 3.0

Theses also provide a useful reference point for documenting the expected keys for various things, like `view_helpers` for helper factories etc and `view_helper_config` for helper specific options.

Also see #110 for additional context

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,12 @@
             "dealerdirect/phpcodesniffer-composer-installer": true
         }
     },
+    "extra": {
+        "laminas": {
+            "component": "Laminas\\View",
+            "config-provider": "Laminas\\View\\ConfigProvider"
+        }
+    },
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
         "ext-dom": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "adb0342243aa19eb2796daf3fd170b5a",
+    "content-hash": "0c0fa8e97afa85daf93f2da903cd926e",
     "packages": [
         {
             "name": "container-interop/container-interop",
@@ -6212,5 +6212,5 @@
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.2.0"
+    "plugin-api-version": "2.3.0"
 }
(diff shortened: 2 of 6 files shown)


froschdesign:
> Having them at least present and setup in composer.json for the component installer means that consumers will get used to having the provider/module injected into app config…

Which is not needed at the moment for an application based on laminas-mvc and a Mezzio application which [uses mezzio-laminasviewrenderer](https://github.com/mezzio/mezzio-laminasviewrenderer/blob/df463cc65105f54418ce2dfa44be6b9918e7a45b/composer.json#L25-L29). :thinking:

gsteel:
@froschdesign This pull is about being able to clean things up in future, and as we go with future pull requests. As we introduce factories for various "things" over time, they can be added to the config provider, once it exists. In future, everything in ViewHelperManager can be removed from instance properties into the config provider providing a clean reference point so that consumers can easily figure out what is registered and how it should be constructed. It also allows us to say, for new projects in MVC and Mezzio, or when updating, inject the config providers to make the migration path to 3.0 easier when it comes.

I'm well aware that the mezzio package does not require this stuff, but in my mind, I believe that certain things currently in the mezzio bridge can be brought into view and removed from there…

froschdesign:
@gsteel 
Everything correct but this not change the fact that module class and the config provider are empty and not needed at the moment, so the pull request can go to a later version than 2.21 or marked as draft.
Or is there something we can add here for version 2.21?

froschdesign:
> Theses also provide a useful reference point for documenting the expected keys for various things, like `view_helpers` for helper factories etc and `view_helper_config` for helper specific options.

You and I might read the code, but the rest look at the documentation. 😉 
This must be added to the documentation.

gsteel:
@froschdesign
I think I understand your reluctance to add the specifically _empty_ config provider/module. My reasoning is also about attempting to keep pull requests small and focussed - this could easily be added during another pull but it's a lot easier to reason about in isolation. Additionally, in answer to adding in 2.21.x… there are _probably_ forthcoming pulls that can make use of this, maybe not 2.21, but certainly 2.x and most definitely 3.x after a merge-up.

I also appreciate the docs comment too… Having the many various config options "documented" in code in the config provider helps those writing real docs to more easily define exactly how and where configuration options should be changed

gsteel:
Closing this PR - it's stale and of limited value right now particularly because work has started on https://github.com/laminas/laminas-mvc-view

PR #164, opened 2022-06-10
Filter Chain Plugin

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| RFC           | yes

### Description

An initial response to an idea proposed by @froschdesign in https://github.com/laminas/laminas-i18n-phone-number/pull/1#discussion_r841077014

### DIFF
--- a/src/Helper/StringFilter.php
+++ b/src/Helper/StringFilter.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\View\Helper;
+
+use Laminas\Filter\FilterChain;
+use Laminas\Filter\FilterInterface;
+use Laminas\Filter\FilterPluginManager;
+use Laminas\View\Exception\InvalidArgumentException;
+
+use function assert;
+use function sprintf;
+
+final class StringFilter
+{
+    private ?FilterChain $chain;
+    private ?string $value;
+    private FilterPluginManager $filterPluginManager;
+
+    public function __constuct(FilterPluginManager $filterPluginManager)
+    {
+        $this->filterPluginManager = $filterPluginManager;
+    }
+
+    public function __invoke(string $value): self
+    {
+        $this->value = $value;
+        $this->chain = new FilterChain();
+
+        return $this;
+    }
+
+    public function __call(string $name, array $arguments)
+    {
+        $this->chain->attach($this->getPlugin($name, $arguments));
+    }
+
+    public function __toString(): string
+    {
+        $value = $this->chain->filter($this->value);
+
+        $this->value = null;
+        $this->chain = null;
+
+        return $value;
+    }
+
+    private function getPlugin(string $name, array $options): FilterInterface
+    {
+        if (! $this->filterPluginManager->has($name)) {
+            throw new InvalidArgumentException(sprintf(
+                'No filter exists with the name "%s"',
+                $name
+            ));
+        }
+
+        $plugin = $this->filterPluginManager->get($name, $options);
+        assert($plugin instanceof FilterInterface);
+
+        return $plugin;
+    }
+}


src/Helper/StringFilter.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function __invoke(string $value): self
+    {
+        $this->value = $value;
+        $this->chain = new FilterChain();
+
+        return $this;
+    }
+
+    public function __call(string $name, array $arguments)
```
Ocramius, line 34: :eyes: :fire:

gsteel: 😂

froschdesign: I understand @Ocramius here and I never use the method either. But here the main goal is ease of use.
Yes, I have the competition in mind, but I want us to do better and still make it as easy as possible to use.
Therefore, we have to find a clever idea to meet all positions.

In the current implementation, the DocBlocks for all available filters are missing. This means some work that could possibly be automated with a script, maybe even with laminas-code. In laminas-validator is [a script to update the hostname validator](https://github.com/laminas/laminas-validator/blob/4875d4e58b6f728981bb767a60530540f82ee1df/bin/update_hostname_validator.php#L1), so this can also be an option here.
But if we use a script for the DocBlocks, then we can also use the script to create the concrete methods for filters in the class immediately. This means that a magical method is no longer necessary.

_We would have to discuss if this would be a practicable way at all and if other problems would not arise as a result!_

gsteel: There's all kinds of things I hate about what I did here, the magic `__call`, the internal state, even the typo `__contruct`, so I'm really happy for it to be shredded, but I agree with @froschdesign that from an ergonomics perspective, using filters to transform content in views would be convenient and useful.

Perhaps a better approach would be to use a known format to build the chain - everyone is familiar with aliases, so maybe something stateless like this…??

`$this->filter($value, 'stringToLower|stringReverse|ucFirst|upsideDown');`

That way, users would be free to use their own custom filters along with those shipped and there would be much less to maintain - just `__construct` and `__invoke` and no state beyond consuming the plugin manager.

Ocramius: > from an ergonomics perspective, using filters to transform content in views would be convenient and useful.

If that's done through a magic call, it's not ergonomic.

 > everyone is familiar with aliases

Probably just me, but stopped using them, since PHP 5.5 and the `::class` syntax made them redundant/unnecessary. 

I don't have full context of the patch, I just saw a **BIG** red flag while scanning the code, hence my original post here

gsteel: > Probably just me, but stopped using them, since PHP 5.5 and the `::class` syntax made them redundant/unnecessary.

Absolutely… second argument could be `list<class-string|string>` to enable FQCNs

> I just saw a BIG red flag

Totally agree - I'm not expecting this to be merged. I opened it here from the phone number lib to keep the idea alive, because whilst I'm not likely to use this myself, I can see how it could be really useful to others.

froschdesign: > … I'm not expecting this to be merged. 

I already marked this pull request as draft.

> Absolutely… second argument could be `list<class-string|string>` to enable FQCNs

I don't see aliases as useful because we need automatic code completion for the filters without knowing the classes themselves or the filter names.

Something like: (pseudo code)

```php
class StringFilter
{
    public function toLower($value): self { … }

    public function reverse($value): self { … }

    public function ucFirst($value): self { … }

    // …

    public function customFilter($value, $className): self { … }

    public function __toString(): string { … }
}

```

gsteel:
Closing this because I have enough to do working towards 3.0 - perhaps we can re-visit once 3.0 is out?

PR #165, opened 2022-07-08
Improve type inference for Plugin Managers

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

General type inference improvements

There's a couple of outstanding psalm issues that I think should probably be baselined.

- HeadLink - psalm now complains about stdClass, which can be annotated as object for the most part, but HeadLink has parameter types explicitly set to `stdClass`
- Psalm is moaning about the plugin managers with `MethodSignatureMismatch` for `Method Laminas\ServiceManager\ServiceManager::has with return type '' is different to return type 'bool'` - This is really weird because bool is the return type all the way up the inheritance chain 🤷‍♂️
- Navigation is looking for non-existent classes - should probably sort that out separately

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -5,6 +5,7 @@
         "intl"
     ],
     "ignore_php_platform_requirements": {
+        "8.0": true,
         "8.1": true
     }
 }
--- a/.psr-container.php.stub
+++ b/.psr-container.php.stub
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psr\Container {
+    /**
+     * Provides automatic type inference for Psalm when retrieving a service from a container using a FQCN
+     */
+    interface ContainerInterface
+    {
+        /** @param string|class-string $id */
+        public function has(string $id): bool;
+
+        /**
+         * @template T of object
+         * @psalm-param string|class-string<T> $id
+         * @psalm-return ($id is class-string ? T : mixed)
+         */
+        public function get(string $id): object;
+    }
+}
--- a/.psr-container.php.stub
+++ b/.psr-container.php.stub
@@ -12,7 +12,7 @@ namespace Psr\Container {
         public function has(string $id): bool;
 
         /**
-         * @template T of object
+         * @template T
          * @psalm-param string|class-string<T> $id
          * @psalm-return ($id is class-string ? T : mixed)
          */
--- a/.psr-container.php.stub
+++ b/.psr-container.php.stub
@@ -8,14 +8,17 @@ namespace Psr\Container {
      */
     interface ContainerInterface
     {
-        /** @param string|class-string $id */
-        public function has(string $id): bool;
+        /**
+         * @param string|class-string $id
+         * @return bool
+         */
+        public function has(string $id);
 
         /**
          * @template T
          * @psalm-param string|class-string<T> $id
          * @psalm-return ($id is class-string ? T : mixed)
          */
-        public function get(string $id): object;
+        public function get(string $id);
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "46c472e2fc0fe89dcb9d1e8ef522dbb4",
+    "content-hash": "f9740d25a5b9d0482c3dedf4adc31831",
     "packages": [
         {
             "name": "laminas/laminas-escaper",
@@ -3305,16 +3305,16 @@
         },
         {
             "name": "phpstan/phpdoc-parser",
-            "version": "1.6.4",
+            "version": "1.5.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpdoc-parser.git",
-                "reference": "135607f9ccc297d6923d49c2bcf309f509413215"
+                "reference": "981cc368a216c988e862a75e526b6076987d1b50"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpdoc-parser/zipball/135607f9ccc297d6923d49c2bcf309f509413215",
-                "reference": "135607f9ccc297d6923d49c2bcf309f509413215",
+                "url": "https://api.github.com/repos/phpstan/phpdoc-parser/zipball/981cc368a216c988e862a75e526b6076987d1b50",
+                "reference": "981cc368a216c988e862a75e526b6076987d1b50",
                 "shasum": ""
             },
             "require": {
@@ -3324,7 +3324,6 @@
                 "php-parallel-lint/php-parallel-lint": "^1.2",
                 "phpstan/extension-installer": "^1.0",
                 "phpstan/phpstan": "^1.5",
-                "phpstan/phpstan-phpunit": "^1.1",
                 "phpstan/phpstan-strict-rules": "^1.0",
                 "phpunit/phpunit": "^9.5",
                 "symfony/process": "^5.2"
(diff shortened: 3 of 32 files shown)


gsteel:
phpcs is broken … conflicting with "phpstan/phpdoc-parser": ">=1.6" solves the issue - not sure if the problem is with the doc parser or slevomat - I'm guessing that this is going to affect anything using coding standard v2.3…

Ocramius:
> * HeadLink - psalm now complains about stdClass, which can be annotated as object for the most part, but HeadLink has parameter types explicitly set to `stdClass`

If we use `@psalm-param object`, it takes priority over any native types

> * Psalm is moaning about the plugin managers with `MethodSignatureMismatch` for `Method Laminas\ServiceManager\ServiceManager::has with return type '' is different to return type 'bool'` - This is really weird because bool is the return type all the way up the inheritance chain man_shrugging

Potential regression in psalm? Perhaps we can reduce this into an example on https://psalm.dev/ and report it

> * Navigation is looking for non-existent classes - should probably sort that out separately

Indeed, let's keep it out for now

.laminas-ci.json
```diff
@@ -5,6 +5,7 @@
         "intl"
     ],
     "ignore_php_platform_requirements": {
+        "8.0": true,
         "8.1": true
```
Ocramius, line 9: We should probably no longer do this for either version :thinking:

.psr-container.php.stub
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Psr\Container {
+    /**
+     * Provides automatic type inference for Psalm when retrieving a service from a container using a FQCN
+     */
+    interface ContainerInterface
+    {
+        /** @param string|class-string $id */
+        public function has(string $id): bool;
```
Ocramius: Is it possible that this is the reason why the boolean type mismatch is reported?

composer.lock
```diff
@@ -4,44 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "adb0342243aa19eb2796daf3fd170b5a",
+    "content-hash": "46c472e2fc0fe89dcb9d1e8ef522dbb4",
     "packages": [
-        {
-            "name": "container-interop/container-interop",
-            "version": "1.2.0",
```
Ocramius, line 11: Gone for good, yay!

src/Helper/Navigation/PluginManager.php
```diff
@@ -16,7 +16,9 @@
  * Navigation\HelperInterface. Additionally, it registers a number of default
  * helpers.
  *
+ * @template InstanceType of HelperInterface|AbstractHelper
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
+ * @extends HelperPluginManager<InstanceType>
  */
 class PluginManager extends HelperPluginManager
```
Ocramius, line 23: We should expand the baseline with this newly reported error

Ocramius (changes requested):
Let's expand the baseline with the new issues: easier to discuss them on the `psalm-baseline.xml` diff, than within comments in the PR

.psr-container.php.stub
gsteel: Oooh - I'll take a look 👍

gsteel: Can confirm that updating the stub solved those issues 🎉

.laminas-ci.json
gsteel: I agree with that, but we'd have to get 8.0 and 8.1 supported in laminas-cli to do that wouldn't we?

Ocramius: Is laminas-cli even involved here?

I think `laminas-console` is still in here - we could update that to be compliant, since it's probably less work than releasing a new major version here :thinking:

gsteel: Sorry, yes, I meant `laminas-console`

gsteel: I just had a quick glance at `laminas-console` - the last abandoned pull request was getting it [up to scratch on 8.0](https://github.com/laminas/laminas-console/pull/15) - it's also been archived. Can we live with ignoring platform reqs, bearing in mind that in 3.0, which does not include console, CI is passing??

gsteel:
@Ocramius - Current phpcs failures only happen on 7.4 - is it possible to run phpcs on 8.0 or 8.1 instead? It would save adding a conflict or pinning versions (slevomat or phpdoc-parser)

.laminas-ci.json
Ocramius: Yes, let's leave it be then :+1:

Ocramius:
> is it possible to run phpcs on 8.0 or 8.1 instead?

We can drop PHP 7.4 support perhaps: I really don't mind on my end, but I'm fairly sure that more people need to have a say in this.

Ocramius:
```
Running ./vendor/bin/phpcs -q --report=checkstyle | cs2pr
PHP Fatal error:  Uncaught TypeError: Argument 5 passed to SlevomatCodingStandard\Helpers\Annotation\ParameterAnnotation::__construct() must be an instance of PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode or null, instance of PHPStan\PhpDocParser\Ast\PhpDoc\TypelessParamTagValueNode given, called in /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationHelper.php on line 357 and defined in /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/Annotation/ParameterAnnotation.php:31
Stack trace:
#0 /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationHelper.php(357): SlevomatCodingStandard\Helpers\Annotation\ParameterAnnotation->__construct('@param', 265, 267, '$mode', Object(PHPStan\PhpDocParser\Ast\PhpDoc\TypelessParamTagValueNode))
#1 /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/SniffLocalCache.php(42): SlevomatCodingStandard\Helpers\AnnotationHelper::SlevomatCodingStandard\Helpers\{clos in /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/Annotation/ParameterAnnotation.php on line 31
```

This looks like a BC break in PHPStan-doc-parser perhaps? /cc @kukulich

kukulich:
@Ocramius Yes, something like this: https://github.com/phpstan/phpdoc-parser/pull/127

It's fixed in Slevomat CS ^8.0

Ocramius:
Hmm, upgrading to next slevomat CS is most likely BC breaking all over the place, so that will take a lot of time :thinking: 

Oh well, renovate-bot **should** be able to help on that.

kukulich:
Locking `phpstan/phpdoc-parser` to 1.5.1 should probably work as well.

Ocramius:
That's a viable solution for upstream, I suppose :thinking:

gsteel:
I added a conflict for 1.6.x which fixed things locally for me - it's just remembering to remove it again in future… I'll add it to the patch

Ocramius:
I think we can update `composer.lock` (only) and stick to that, for now.

gsteel:
> I think we can update `composer.lock` (only) and stick to that, for now.

How do you change deps in the lock without modifying composer.json?

Ocramius:
1. add the `"the-dependency": "1.2.3"` to `composer.json`
2. run `composer update "the-dependency"`
3. remove `"the-dependency"` from `composer.json`
4. run `composer update nothing`

gsteel:
Thanks @Ocramius - I finally got there 🙄

Ocramius (approved):
:ship: this brings us forward!

PR #170, opened 2022-07-18
Correct plugin manager type annotations

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Corrects the type annotation for the Helper Plugin Manager.

The plugin manager accepts and returns either implementations of HelperInterface or a callable - the previous types were incorrect.

2 psalm issues are baselined for `HelperPluginManager::get()`, `MixedInferredReturnType` and `MixedReturnStatement` because psalm cannot guarantee that get() does not return mixed as defined by AbstractPluginManager when given an arbitrary string.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1594,7 +1594,7 @@
       <code>$container</code>
     </MixedAssignment>
     <MixedInferredReturnType occurrences="1">
-      <code>($name is class-string ? T : callable)</code>
+      <code>($name is class-string ? T : HelperInterface|callable)</code>
     </MixedInferredReturnType>
     <MixedMethodCall occurrences="8">
       <code>get</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1593,6 +1593,9 @@
       <code>$container</code>
       <code>$container</code>
     </MixedAssignment>
+    <MixedInferredReturnType occurrences="1">
+      <code>($name is class-string ? T : callable)</code>
+    </MixedInferredReturnType>
     <MixedMethodCall occurrences="8">
       <code>get</code>
       <code>get</code>
@@ -1603,6 +1606,9 @@
       <code>has</code>
       <code>has</code>
     </MixedMethodCall>
+    <MixedReturnStatement occurrences="1">
+      <code>parent::get($name, $options)</code>
+    </MixedReturnStatement>
     <PropertyTypeCoercion occurrences="1">
       <code>$this-&gt;initializers</code>
     </PropertyTypeCoercion>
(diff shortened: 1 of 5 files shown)


psalm-baseline.xml
```diff
@@ -1593,6 +1593,9 @@
       <code>$container</code>
       <code>$container</code>
     </MixedAssignment>
+    <MixedInferredReturnType occurrences="1">
+      <code>($name is class-string ? T : callable)</code>
```
Ocramius: I think this should be `: HelperInterface|callable`, not just `callable`

Ocramius (changes requested):

psalm-baseline.xml
gsteel: This is a bit of a tough one I think. Falling back to the union type there makes whatever you retrieve impossible to call without asserting it's either one or the other because HelperInterface isn't invokable. I think I went for `? T : callable` because that gives users `...->get('myThing')->__invoke()` without an error which is the most likely use-case no?

test/StaticAnalysis/PluginRetrieval.php
```diff
@@ -0,0 +1,38 @@
...
+    public function retrieveHelperByClassName(string $email): string
+    {
+        return ($this->pluginManager->get(GravatarImage::class))($email);
+    }
+
+    public function retrievalByAliasAssumesCallable(): void
+    {
+        $helper = $this->pluginManager->get('something');
+        $helper();
+    }
```
gsteel: This is broken if get() is annotated with `($id is class-string ? T : HelperInterface|callable)`

psalm-baseline.xml
Ocramius: Careful: `callable` does not mean `->__invoke()`, since an `array` is a valid callable sometimes :-) 

If you do `->get('foo')`, it may also be a helper

test/StaticAnalysis/PluginRetrieval.php
Ocramius: Yep, that seems normal

psalm-baseline.xml
gsteel: Absolutely - you are completely right, I just incorrectly considered inferring the return type as any type of callable would be more useful for consumers even though it's less precise. 👍

psalm-baseline.xml
```diff
@@ -1593,6 +1593,9 @@
       <code>$container</code>
       <code>$container</code>
     </MixedAssignment>
+    <MixedInferredReturnType occurrences="1">
+      <code>($name is class-string ? T : HelperInterface|callable)</code>
+    </MixedInferredReturnType>
```
Ocramius, line 1598: Wait, psalm can't work with this? :thinking:

gsteel: No, because parent::get() returns `? T : mixed`, but in this case we _know_ that it's a callable or HelperInterface.

I could add an assertion, or a second call to `$this->validate()` ??

Ocramius: Ah, it cannot infer the return type, but the declared one is still valid for consumers, right?

gsteel: Yes!

Ocramius (approved):
LGTM :+1:

PR #173, opened 2022-08-17
Removes asserting an implementation from the Layout helper

### LINKED ISSUE #172: A new assertion added into Layout prohibits use of OSS renderers such as provided by Twig.  Would like to soften it.
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | somewhat

#### Summary

An assertion was added to Layout::getViewModelHelper() which prohibits OSS implementations of RendererInterface in the wild, for example, Twig's [TwigRenderer](https://github.com/kokspflanze/ZfcTwig/blob/master/src/View/TwigRenderer.php).

    protected function getViewModelHelper()
    {
        if (! $this->viewModelHelper) {
            /**
             * @psalm-suppress DeprecatedMethod
             */
            $renderer = $this->getView();
            assert($renderer instanceof PhpRenderer); // offending assertion
            $helper = $renderer->plugin('view_model');
            assert($helper instanceof ViewModel);
            $this->viewModelHelper = $helper;
        }

        return $this->viewModelHelper;
    }

Adding plugin() to RendererInterface seems a bit heavy handed, but relaxing the assertion to an interface would seem to be more logical, specially since the assertion only serves to protect the contract for ->plugin

#### Proposal

Would you be open to accepting a PR that either proposes to substitute this assertion for one that looks for an interface that provides plugin() (couldn't find an existing one), or one that adds `plugin` to RendererInterface (implemented by PhpRenderer)?

Saeven:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | somewhat

Fixes #172 

#### Summary

An assertion was added to Layout::getViewModelHelper() which prohibits OSS implementations of RendererInterface in the wild, for example, Twig's [TwigRenderer](https://github.com/kokspflanze/ZfcTwig/blob/master/src/View/TwigRenderer.php).

    protected function getViewModelHelper()
    {
        if (! $this->viewModelHelper) {
            /**
             * @psalm-suppress DeprecatedMethod
             */
            $renderer = $this->getView();
            assert($renderer instanceof PhpRenderer); // offending assertion
            $helper = $renderer->plugin('view_model');
            assert($helper instanceof ViewModel);
            $this->viewModelHelper = $helper;
        }

        return $this->viewModelHelper;
    }

### DIFF
--- a/src/Helper/Layout.php
+++ b/src/Helper/Layout.php
@@ -100,7 +100,7 @@ class Layout extends AbstractHelper
     {
         if (! $this->viewModelHelper) {
             /**
-             * @psalm-suppress DeprecatedMethod
+             * @psalm-suppress PossiblyNullReference, UndefinedInterfaceMethod
              */
             $helper = $this->getView()->plugin('view_model');
             assert($helper instanceof ViewModel);
--- a/src/Helper/Layout.php
+++ b/src/Helper/Layout.php
@@ -99,9 +99,10 @@ class Layout extends AbstractHelper
     {
         if (! $this->viewModelHelper) {
             /**
-             * @psalm-suppress PossiblyNullReference, UndefinedInterfaceMethod
+             * @psalm-suppress DeprecatedMethod
              */
-            $helper = $this->getView()->plugin('view_model');
+            $renderer = $this->getView();
+            $helper = $renderer->plugin('view_model');
             assert($helper instanceof ViewModel);
             $this->viewModelHelper = $helper;
         }
--- a/src/Helper/Layout.php
+++ b/src/Helper/Layout.php
@@ -99,7 +99,7 @@ class Layout extends AbstractHelper
     {
         if (! $this->viewModelHelper) {
             $renderer = $this->getView();
-            $helper = $renderer->plugin('view_model');
+            $helper   = $renderer->plugin('view_model');
             assert($helper instanceof ViewModel);
             $this->viewModelHelper = $helper;
         }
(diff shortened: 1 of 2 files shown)


src/Helper/Layout.php
```diff
@@ -100,11 +100,9 @@ protected function getViewModelHelper()
     {
         if (! $this->viewModelHelper) {
             /**
-             * @psalm-suppress DeprecatedMethod
+             * @psalm-suppress PossiblyNullReference, UndefinedInterfaceMethod
```
Ocramius: I think we shouldn't suppress this, but rather baseline it: the problem is there, but we just didn't fix it yet.

Ocramius (changes requested):

src/Helper/Layout.php
Saeven: Yeah was just trying to keep the CI chain happy.  You're ok with a psalm break?

Ocramius: Nono, just regenerate the baseline file 👍

Ocramius (changes requested):
The psalm baseline didn't get expanded?

Saeven:
Yeah sorry, gimme a sec while I work it out - psalm being contentious on my machine for some reason (only in this project, well behaved on company code)

Saeven:
Generating blank files for some reason, but I added them by hand.

Ocramius (approved):
Thanks @Saeven!

PR #193, opened 2023-01-14
Fix psalm issues due to improved upstream types

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

- Bump coding standard to 2.5
- Bump psalm to 5.4
- Update (some) locked deps

`laminas-console` is beginning to make it impossible to keep dependencies sane - Ignoring platform reqs is becoming more unsafe on PHP 8.0 for example.

Closes #192

PR #194, opened 2023-01-14
Remove laminas/console dev dependency and stop ignoring platform requirements

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Includes commits from #193 - Removes console so that deps can be safely updated again. See failures for 8.0 latest in #193


Ocramius (approved):
Thanks @gsteel!

PR #198, opened 2023-02-08
Fixed improper navigation semantics in accessibility contexts

### LINKED ISSUE #197: Improper navigation semantics
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | `^2.8`

#### Summary

When a link to the currently displayed page is rendered, it should include the `aria-current="page"` attribute.

#### Current behavior

Navigation links are not providing "current page" semantics in the accessibility tree.

#### How to reproduce

Using the screen reader of your choice, activate a menu link, then bring focus back to that same a menu element.

#### Expected behavior

Varies based on assistive technology being used. As an example, VoiceOver macOS will read a properly tagged navigation link as "current page, visited, link, Example Page".

ivyrze:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Adds an `aria-current="page"` to active menu links and breadcrumb segments, resolving #197.

The menu tests as previously designed have more than one page marked as active, which means `aria-current` is being used improperly in those test scenarios. I'm assuming that in real-world use, this would not be the case?

### DIFF
--- a/src/Helper/Navigation/Breadcrumbs.php
+++ b/src/Helper/Navigation/Breadcrumbs.php
@@ -111,9 +111,13 @@ class Breadcrumbs extends AbstractHelper
         } else {
             /** @var View\Helper\EscapeHtml $escaper */
             $escaper = $this->view->plugin('escapeHtml');
-            $html    = $escaper(
+            $label   = $escaper(
                 $this->translate($active->getLabel(), $active->getTextDomain())
             );
+            $attribs = [
+                'aria-current' => 'page',
+            ];
+            $html    = '<span' . $this->htmlAttribs($attribs) . '>' . $label . '</span>';
         }
 
         // walk back to root
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.4.0@62db5d4f6a7ae0a20f7cc5a4952d730272fc0863">
+<files psalm-version="5.6.0@e784128902dfe01d489c4123d69918a9f3c1eac5">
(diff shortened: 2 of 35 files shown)


src/Helper/Navigation/Breadcrumbs.php
```diff
@@ -111,9 +111,13 @@ public function renderStraight($container = null)
         } else {
             /** @var View\Helper\EscapeHtml $escaper */
             $escaper = $this->view->plugin('escapeHtml');
-            $html    = $escaper(
+            $label   = $escaper(
```
Ocramius, line 114: The type seems valid here: re-generating the Psalm baseline should suffice. Not sure why `mixed` is inferred, but your code is good as-is :+1:

src/Helper/Navigation/Breadcrumbs.php
```diff
@@ -111,9 +111,13 @@ public function renderStraight($container = null)
...
                 $this->translate($active->getLabel(), $active->getTextDomain())
             );
+            $attribs = [
+                'aria-current' => 'page',
+            ];
+            $html    = '<span' . $this->htmlAttribs($attribs) . '>' . $label . '</span>';
```
Ocramius, line 120: Similar to above: re-generating the psalm baseline will suffice

Ocramius (approved):
Besides the small Psalm misalignment, this LGTM! 

Thanks @ivyrze. Give the baseline a try, and if it fails, I'll pick it from here, just let me know.

ivyrze:
Thanks for the pointer @Ocramius!

Ocramius (approved):
Thanks @ivyrze!

PR #200, opened 2023-02-17
fix(php-doc): set good param-out typing to reflect code behavior

neilime:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

`Laminas\View\Helper\Navigation\AbstractHelper::parseContainer` method may set the `$container` param to  `AbstractContainer` or null.

But the PHPDoc says it set always to `AbstractContainer`.

It leads to code analyses issues. Example in `Breadcrumbs`:

```php
    public function renderStraight($container = null)
    {
        $this->parseContainer($container);
        if (null === $container) {
            $container = $this->getContainer();
        }
```

> Strict comparison using === between null and Laminas\Navigation\AbstractContainer will always evaluate to false.

### DIFF
--- a/src/Helper/Navigation/AbstractHelper.php
+++ b/src/Helper/Navigation/AbstractHelper.php
@@ -55,7 +55,7 @@ abstract class AbstractHelper extends View\Helper\AbstractHtmlElement implements
     /**
      * AbstractContainer to operate on by default
      *
-     * @var AbstractContainer
+     * @var AbstractContainer|null
      */
     protected $container;
 
@@ -188,6 +188,10 @@ abstract class AbstractHelper extends View\Helper\AbstractHtmlElement implements
     public function findActive($container, $minDepth = null, $maxDepth = -1)
     {
         $this->parseContainer($container);
+        if (null === $container) {
+            return [];
+        }
+
         if (! is_int($minDepth)) {
             $minDepth = $this->getMinDepth();
         }
@@ -244,7 +248,7 @@ abstract class AbstractHelper extends View\Helper\AbstractHtmlElement implements
      *
      * @param AbstractContainer|string|null $container
      * @return void
-     * @param-out AbstractContainer $container
+     * @param-out AbstractContainer|null $container
      * @throws Exception\InvalidArgumentException
      */
     protected function parseContainer(&$container = null)


src/Helper/Navigation/AbstractHelper.php
```diff
@@ -244,7 +244,7 @@ public function findActive($container, $minDepth = null, $maxDepth = -1)
      *
      * @param AbstractContainer|string|null $container
      * @return void
-     * @param-out AbstractContainer $container
+     * @param-out AbstractContainer|null $container
```
Ocramius, line 247: Possibly a good idea to declare a parametrized `param-out`, to remove nullability scenarios from a call-site perspective

Ocramius, line 251: Possibly a good idea to declare a parametrized `param-out`, to remove nullability scenarios from a call-site perspective

Ocramius (changes requested):

src/Helper/Navigation/AbstractHelper.php
```diff
@@ -188,6 +188,10 @@ public function __toString()
     public function findActive($container, $minDepth = null, $maxDepth = -1)
     {
         $this->parseContainer($container);
+        if (null === $container) {
+            return [];
+        }
```
Ocramius, line 193: This code change requires a corresponding test case addition

Ocramius (changes requested):

src/Helper/Navigation/AbstractHelper.php
neilime: Sadly it looks not possible with PSALM: https://github.com/vimeo/psalm/issues/8176

src/Helper/Navigation/AbstractHelper.php
neilime: Actually this is complicated to add test for this.
Because we technically can pass null as $container to `findActive` method, but we cannot create a test for this because PSALM will says "Argument 1 of Laminas\View\Helper\Navigation\Breadcrumbs::findActive cannot be null"

As PSALM will not support conditional param-out, it is the only way I've found to satisfy both code behavior and PSALM checks (Example: https://github.com/laminas/laminas-view/actions/runs/4206486056)

src/Helper/Navigation/AbstractHelper.php
Ocramius: Then let's skip until possible, IMO

Ocramius (changes requested):
Let's leave this until better `param-out` can be used.

Alternatively, deprecate the method entirely, and have a proper method with input and output values instead (no by-ref)

neilime:
I will not refactor a whole class just because of wrong typing / poor capabilities of PSALM.
Let's  wait then...

Ocramius:
Closing here meanwhile.

PR #205, opened 2023-05-29
Remove support for PHP 8.0. Bump ServiceManager to 3.21.x

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| QA            | yes

### Description

Includes commits from #201

Works around a BC break relating to tests introduced in `laminas-servicemanager` 3.21 where a protected method was made static.

Resolves SA issues introduced by the PHP upgrade and baselines other (currently) unfixable SA issues

### DIFF
--- a/src/Helper/Navigation/AbstractHelper.php
+++ b/src/Helper/Navigation/AbstractHelper.php
@@ -434,7 +434,7 @@ abstract class AbstractHelper extends View\Helper\AbstractHtmlElement implements
     protected function translate($message, $textDomain = null)
     {
         if (! is_string($message) || empty($message)) {
-            return '';
+            return $message;
         }
 
         if (! $this->isTranslatorEnabled() || ! $this->hasTranslator()) {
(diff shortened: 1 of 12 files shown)


src/Helper/Navigation/AbstractHelper.php
```diff
@@ -432,14 +434,15 @@ public function htmlify(AbstractPage $page)
     protected function translate($message, $textDomain = null)
     {
         if (! is_string($message) || empty($message)) {
-            return $message;
+            return '';
```
Ocramius: BC issue?

Ocramius (changes requested):

src/Helper/Navigation/AbstractHelper.php
gsteel: Yes, you're right of course - I've reverted and baselined the errors. Thanks for the review @Ocramius

froschdesign (approved):

Ocramius (approved):
Thanks @gsteel!

PR #207, opened 2023-05-31
Upgrade PHPUnit to 10.1.x

gsteel:
Closes #196

### DIFF
--- a/test/Helper/Navigation/AbstractTestCase.php
+++ b/test/Helper/Navigation/AbstractTestCase.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace LaminasTest\View\Helper\Navigation;
 
 use Laminas\I18n\Translator\Translator;
+use Laminas\ModuleManager\ModuleManager;
+use Laminas\Mvc\Application;
 use Laminas\Mvc\Service\ServiceManagerConfig;
 use Laminas\Navigation\Navigation;
 use Laminas\Navigation\Service\DefaultNavigationFactory;
@@ -30,7 +32,7 @@ use function file_get_contents;
  *
  * @psalm-suppress MissingConstructor
  */
-abstract class AbstractTest extends TestCase
+abstract class AbstractTestCase extends TestCase
 {
     /** @var ServiceManager */
     protected $serviceManager;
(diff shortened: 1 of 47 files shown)


test/Helper/Navigation/AbstractTestCase.php
```diff
@@ -175,7 +179,7 @@ protected function getAcl(): array
 
         $acl->addResource(new GenericResource('guest_foo'));
         $acl->addResource(new GenericResource('member_foo'), 'guest_foo');
-        $acl->addResource(new GenericResource('admin_foo', 'member_foo'));
+        $acl->addResource(new GenericResource('admin_foo'));
```
froschdesign, line 182: It looks like it doesn't matter, but I think that was the original intention:
```suggestion
        $acl->addResource(new GenericResource('admin_foo'), 'member_foo');
```

gsteel: Yes it was, but the string fixtures being compared against fail when `admin_foo` is given a parent role, so I just deleted the extra argument

froschdesign: I mean the difference between the brackets that leads to different arguments:

```php
$acl->addResource(new GenericResource('admin_foo', 'member_foo'));
// vs.
$acl->addResource(new GenericResource('admin_foo'), 'member_foo');
```

froschdesign: See also: https://github.com/laminas/laminas-view/pull/207/files#diff-3208225af5a3ef0c924761633ad6b04cbfa5eab3d3bafcc49c79441e89257ee7L2692-L2694

gsteel: Acknowledged. I am aware of the original mistake. Removing the offending and redundant parameter affects nothing, but correcting the mistake yields different navigation structures due to the change to the ACL configuration, which in turn causes tests to fail needlessly. i.e. it's not worth correcting the misplaced parenthesis.

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @gsteel!

I don't think a release is needed here though, as nothing user-facing was merged.

PR #209, opened 2023-07-19
Clean up HeadStyle View Helper

### LINKED ISSUE #6: Inline CSS always commented by HeadStyle ViewHelper
It looks like the inline style definition is not correctly printed by HeadStyle view helper.

Considering the following one:

```php
<?php $this->headStyle()->captureStart(); ?>
@media only screen and (max-width: 620px) {

div[class*="pd-left20"]{padding:0 0 0 20px !important;}
td[class*="resetPadding"]{padding:0 0 0 0 !important;}
td[class*="padding20"]{padding:20px 0 20px 0 !important;}

}
<?php $this->headStyle()->captureEnd(); ?>
```

The whole block will be commented out.

I tried disabling auto escaper but escaping is hardcoded into HeadStyle implementation.

Actually, after a deep look in the code, I don't get why CSS content is always commented via ```<!-- -->```:

[[0]](#0)
```php
 $html = '<style type="text/css"' . $attrString . '>' . PHP_EOL
            . $escapeStart . $indent . $item->content . PHP_EOL . $escapeEnd
            . '</style>';
```

$escapeStart and $escapeEnd can be null only for conditional CSS which is not a regular use case.

The examples from documentation are giving me similar commented output. [[1]](#1)

<a name="0">[0]</a> https://github.com/zendframework/zend-view/blob/master/src/Helper/HeadStyle.php#L328
<a name="1">[1]</a> https://docs.zendframework.com/zend-view/helpers/head-style/#capturing-style-declarations

---

Originally posted by @tworzenieweb at https://github.com/zendframework/zend-view/issues/165

gsteel:
Closes #6

It is not necessary to comment CSS within style tags so output has changed from:

`<style><!-- a {} --></style>` to `<style> a {} </style>`

The `<style media="">` attribute was restricted to list of pre-defined values such as `[screen, print, braille]`, when in reality the media attribute can contain any sort of media query such as `screen and (min-width: 10rem)`. These restrictions have been removed the property listing those restrictions deprecated.

Additionally, if the `media` attribute is an array containing non-string values, an exception will be thrown. Previously, there would have been a type error.

Style tag attributes not present in a list of 'allowed' attributes are silently ignored. This list has been updated to include the `nonce` attribute.

Adds `@method setIndent` which magically proxies to the container.

Adds probably desirable behaviour where if the user provides an empty string, i.e. `$this->headStyle()->appendStyle('');`, then no markup will be output for that item.

Lots of type inference improvements and some test improvements.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.12.0@f90118cdeacd0088e7215e64c0c99ceca819e176">
+<files psalm-version="5.13.1@086b94371304750d1c673315321a55d15fc59015">
   <file src="bin/templatemap_generator.php">
     <MissingParamType>
       <code>$templatePath</code>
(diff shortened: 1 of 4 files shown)


froschdesign (approved):
Nice improvement and long overdue. 👍🏻 

(Yes, the output is changed, but that doesn't matter with testing tools like Playwright or Cypress.)

PR #210, opened 2023-07-19
Improve type inference for stateful container implementations

gsteel:
Adds templates to the container and abstract placeholder view helper classes and documents the types used by each implementing view helper.

A lot of incorrect docblocks have been fixed, additional magic methods documented and a decent reduction in the baseline.

<strike>Several implementations store values as plain objects and were in places (parameter) type hinting on `stdClass`. All references to `stdClass` have been replaced with `object` and `instanceof` conditions replaced with `is_object`. There's no BC break because `object` is more general than `stdClass`. The main purpose here is that it's not possible to express object shapes with `stdClass` which is central to the significant improvements in type inference.</strike>

Methods that expect `stdClass` have been annotated to accept `object{foo:string}` which results in a lot fewer psalm issues and better inference than the handful of `ArgumentTypeCoercion` and `MismatchingDocblockParamType` added to the baseline.

Various public and protected methods have been marked `@internal` so that they can be made private in the next major, and there are a handful of deprecations for unused, un-documented and un-tested public methods.

### DIFF
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -358,17 +358,18 @@ class HeadLink extends AbstractStandalone
      */
     public function toString($indent = null)
     {
-        $indent = null !== $indent
-                ? $this->getContainer()->getWhitespace($indent)
-                : $this->getContainer()->getIndent();
+        $container = $this->getContainer();
+        $indent    = null !== $indent
+                ? $container->getWhitespace($indent)
+                : $container->getIndent();
 
         $items = [];
-        $this->getContainer()->ksort();
-        foreach ($this as $item) {
+        $container->ksort();
+        foreach ($container as $item) {
             $items[] = $this->itemToString($item);
         }
 
-        return $indent . implode($this->escape($this->getContainer()->getSeparator()) . $indent, $items);
+        return $indent . implode($this->escape($container->getSeparator()) . $indent, $items);
     }
 
     /**
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -5,17 +5,17 @@ declare(strict_types=1);
 namespace Laminas\View\Helper;
 
 use Laminas\View\Exception;
-use stdClass;
+use Laminas\View\Helper\Placeholder\Container\AbstractContainer;
+use Laminas\View\Helper\Placeholder\Container\AbstractStandalone;
 
 use function array_intersect;
 use function array_keys;
 use function array_shift;
-use function call_user_func_array;
 use function count;
-use function func_get_args;
 use function get_object_vars;
 use function implode;
 use function is_array;
+use function is_object;
 use function is_string;
 use function method_exists;
 use function preg_match;
@@ -24,11 +24,8 @@ use function str_replace;
 
 use const PHP_EOL;
 
-// @codingStandardsIgnoreStart
 /**
- * Laminas_Layout_View_Helper_HeadLink
- *
- * @see http://www.w3.org/TR/xhtml1/dtds.html
+ * @extends AbstractStandalone<int, object>
  *
  * Creates the following virtual methods:
  * @method HeadLink appendStylesheet($href, $media = 'screen', $conditionalStylesheet = '', $extras = [])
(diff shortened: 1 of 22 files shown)


src/Helper/HeadLink.php
```diff
@@ -201,12 +197,14 @@ public function __call($method, $args)
...
-     * @param  mixed $value
+     * @internal This method will become private in version 3.0
+     *
+     * @param mixed $value
      * @return bool
      */
     protected function isValid($value)
     {
-        if (! $value instanceof stdClass) {
+        if (! is_object($value)) {
```
Ocramius, line 209: Hmm, this seems like a BC break

src/Helper/HeadLink.php
```diff
@@ -357,23 +359,25 @@ public function itemToString(stdClass $item)
     public function toString($indent = null)
     {
         $indent = null !== $indent
-                ? $this->getWhitespace($indent)
-                : $this->getIndent();
+                ? $this->getContainer()->getWhitespace($indent)
+                : $this->getContainer()->getIndent();
```
Ocramius: Let's not `getContainer()` all over the place :D

Ocramius: (use a variable)

src/Helper/HeadLink.php
```diff
@@ -383,8 +387,10 @@ public function createData(array $attributes)
     /**
      * Create item for stylesheet link item
      *
+     * @deprecated This method is unused and will be removed in version 3.0 of this component
+     *
      * @param  array $args
-     * @return stdClass|false Returns false if stylesheet is a duplicate
+     * @return object|false Returns false if stylesheet is a duplicate
```
Ocramius, line 396: IMO we should hold this back a bit - this change is potentially breaking

Ocramius (commented):
Only did a quick skim as I'm on the road, soz

src/Helper/HeadLink.php
gsteel: That's a deprecated, undocumented and internally unused method and it still actually returns a `stdClass`. For HeadLink specifically, the internal type is just `(object) array<string, mixed>`, in other implementations, documented return types have been changed to Psalm object shapes. I can't see any practical way that BC is broken here.

Returning `object` here just satisfies the template for `TValue`.

Can you explain how this might be a BC break?

src/Helper/HeadLink.php
gsteel: For users (inheritors), behaviour hasn't changed - if anything, it's a little more relaxed now as you could more easily provide a value object, providing it has the expected public properties. All of these similar `isValid` methods do the same thing: Is it an object and does it have the properties required? No behaviour has changed.

gsteel:
> Only did a quick skim as I'm on the road, soz

Thanks for looking at it @Ocramius

src/Helper/HeadLink.php
```diff
@@ -290,15 +288,19 @@ public function set($value)
...
 
     /**
      * Create HTML link element from data item
      *
+     * @internal This method will become private in version 3.0
+     *
      * @return string
      */
-    public function itemToString(stdClass $item)
+    public function itemToString(object $item)
```
froschdesign: If I test this locally, it ends in a BC break.

gsteel: Yes. My bad… I'm getting mixed up with covariance and contravariance. This is a BC break for inheritors - I'll revert parameter type changes throughout 👍

src/Helper/HeadLink.php
Ocramius: The expectation downstream is `stdClass`,  not `object`, heh :D

Ocramius (approved):
Went through this in more detail: the move from `stdClass` to `object` is fine, and `stdClass` is kept at runtime anyway, so we're OK here, no BC breaks :+1: 

Thanks @gsteel!

Ocramius:
Nice improvements BTW! 

Selection_166

https://shepherd.dev/github/laminas/laminas-view

FabianKoestring:
Why was `@final` added to some classes?

Ocramius:
@FabianKoestring we'll make many classes `final` in the next major release, since allowing inheritance everywhere has hurt our maintainability in a very substantial way.

PR #212, opened 2023-07-20
Deprecations and type inference improvements for template resolvers

gsteel:
The return type for a template resolver _should_ simply be `string`. Any kind of failure should cause an exception, ideally something like `TemplateNotFound`.

All existing resolvers are marked as `@final` and there are several additional deprecations to clean up un-documented behaviour.

The baseline should be relatively even - most of the solved issues have been replaced with deprecation issues + some new issues surfaced by the type inference improvements.

## Deprecations

- Additional protected properties relating to already deprecated stream wrappers
- public error constants that will be unnecessary in 3.0 when resolvers throw exceptions instead of returning false.
- (Undocumented) Using `TemplateMapResolver::add(string, null)` to remove individual templates - this now triggers `E_USER_DEPRECATED`
- `AggregateResolver::getLastSuccessfulResolver()` - likely exists only for tests
- `AggregateResolver::getLastLookupFailure()` - again, probably only there for tests
- `TemplatePathStack::getLastLookupFailure()` - as above

### DIFF
--- a/src/Resolver/TemplateMapResolver.php
+++ b/src/Resolver/TemplateMapResolver.php
@@ -9,20 +9,26 @@ use IteratorAggregate;
 use Laminas\Stdlib\ArrayUtils;
 use Laminas\View\Exception;
 use Laminas\View\Renderer\RendererInterface as Renderer;
-use ReturnTypeWillChange; // phpcs:ignore
+use ReturnTypeWillChange;
 use Traversable;
 
 use function array_key_exists;
 use function array_replace_recursive;
-use function gettype;
-use function is_array;
-use function is_object;
+use function get_debug_type;
+use function is_iterable;
 use function is_string;
 use function sprintf;
+use function trigger_error;
 
+use const E_USER_DEPRECATED;
+
+/**
+ * @implements IteratorAggregate<string, string>
+ * @final
+ */
 class TemplateMapResolver implements IteratorAggregate, ResolverInterface
 {
-    /** @var array */
+    /** @var array<string, string> */
     protected $map = [];
 
     /**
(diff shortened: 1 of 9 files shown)


src/Resolver/TemplateMapResolver.php
```diff
@@ -78,51 +84,54 @@ public function setMap($map)
...
      */
     public function add($nameOrMap, $path = null)
     {
-        if (is_array($nameOrMap) || $nameOrMap instanceof Traversable) {
-            $this->merge($nameOrMap);
+        if (is_string($nameOrMap) && ($path === null || $path === '')) {
+            trigger_error(
+                'Using add() to remove individual templates is deprecated and will be removed in version 3.0',
+                E_USER_DEPRECATED,
+            );
```
froschdesign, line 98: No roadmap has yet been written down for this, but when these deprecation errors are triggered then this should also be reflected in the documentation:

- Migration
  - Preparing for Version 3

src/Resolver/TemplatePathStack.php
```diff
@@ -72,18 +84,22 @@ class TemplatePathStack implements ResolverInterface
      * Flags used to determine if a stream wrapper should be used for enabling short tags
      */
 
-    /** @var bool */
+    /**
+     * @deprecated Stream wrapper functionality will be removed in version 3.0 of this component
```
froschdesign, line 88: The same as above.

gsteel:
@froschdesign
Will this be good-to-go once #216 is merged?

src/Resolver/TemplateMapResolver.php
gsteel: Handled in #216

froschdesign (approved):

PR #220, opened 2023-10-17
Add return type hint to ViewModel::getIterator()

### LINKED ISSUE #73: Type hint for `ViewModel::getIterator()`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary
To satisfy static code analyzers like `phpstan` or `psalm`, we should add a type hint for `ViewModel::getIterator()` to avoid issues like these: https://phpstan.org/r/4b5b2db4-58c5-45c1-878d-f107d605fd69 (PHPStan example. Line 11 and 13: '...type has no value type specified in iterable type...').

This should suffice:

```
    /**
     * Get iterator of children
     *
     * @return ArrayIterator<int, ModelInterface>
     */
    public function getIterator()
    {
        return new ArrayIterator($this->children);
    }
```

How is your stand on this?
This works with PHPStan, but I know Laminas decided to go with Psalm. Is this suitable for Psalm, too?

func0der:
This is to satisfy static code analytic tools like PHPStan and Psalm.

Resolves #73 and also helps with #56

### DIFF
--- a/src/Model/ViewModel.php
+++ b/src/Model/ViewModel.php
@@ -32,6 +32,7 @@ class ViewModel implements ModelInterface, ClearableModelInterface, RetrievableC
      * Child models
      *
      * @var array
+     * @psalm-type list<ModelInterface>
      */
     protected $children = [];
 
@@ -508,7 +509,7 @@ class ViewModel implements ModelInterface, ClearableModelInterface, RetrievableC
     /**
      * Get iterator of children
      *
-     * @return ArrayIterator
+     * @return Traversable<int, ModelInterface>
      */
     #[ReturnTypeWillChange]
     public function getIterator()


gsteel (changes requested):
Thanks @funcoder 👍

Really, this should be `Traversable<int, ModelInterface>` to fit with the return type of `IteratorAggregate` - when we eventually get return types added in 3.0, This will receive `Traversable` as its native return type.

Also, can you please update and target the latest minor `2.31.x`. Your change is not a BC break so it can be released in a minor.

gsteel:
I should also mention that to keep psalm happy here, you'll need to annotate `$this->children` as `list<ModelInterface>`

func0der:
:roll_eyes: Github interface confused me with the branches xD 

I hope I got everything the way you wanted it to be, @gsteel ?

gsteel (approved):
LGTM! Thanks @func0der

PR #233, opened 2024-01-23
Update Menu.php

tarcisiocms:
The consumption of the findActive function may be conditioned on the $onlyActive variable since the use of the $foundPage and $foundDepth variables within foreach is also.

In applications with many pages, if you do not use the onlyActive option, the return time can be optimized by conditioning the findActive method call

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### DIFF
--- a/src/Helper/Navigation/Menu.php
+++ b/src/Helper/Navigation/Menu.php
@@ -280,7 +280,7 @@ class Menu extends AbstractHelper
             // find deepest active
             $found = $this->findActive($container, $minDepth, $maxDepth);
             if ($found) {
-                $foundPage = $found['page'];
+                $foundPage  = $found['page'];
                 $foundDepth = $found['depth'];
             }
         }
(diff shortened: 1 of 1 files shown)


src/Helper/Navigation/Menu.php
```diff
@@ -270,18 +270,19 @@ protected function renderNormalMenu(
...
-            $foundPage  = $found['page'];
-            $foundDepth = $found['depth'];
+        if ($onlyActive) {
+            // find deepest active
+            $found = $this->findActive($container, $minDepth, $maxDepth);
+            if ($found) {
+                $foundPage  = $found['page'];
+                $foundDepth = $found['depth'];
+            }
         }
```
Ocramius, line 286: Please do add a test to verify these changes

Ocramius (changes requested):

gsteel:
The navigation view helpers are no longer present in 3.0.x and it looks like this patch has been abandoned. Please re-open if it's important to see this land in 2.x

Issue #149, opened 2022-02-23
[RFC]: Massive BC Break and Simplification of `laminas-view`


gsteel:
With the approaching 3.0 release, I'd like to propose a significant simplification of the laminas-view rendering architecture so that it consists of fewer components with smaller apis.

- A Template resolver _(of which there should remain  a number of implementations to satisfy the use case of different resolution mechanisms and directory layouts)_ becomes a hard dependency of the template renderer.
- Remove the concept of rendering strategies
- Remove all renderers but the PHP renderer.
- The helper plugin manager becomes a hard dependency of the concrete renderer.
- Potentially, `View` and `PhpRenderer` become the same thing.

Effectively setting up a complete, fully functioning template renderer would then become:

```php

$templateResolver = new TemplateResolver('/template-directory');
$pluginManager = new ViewHelperManager([/** configuration */]);
$renderer = new PhpRenderer($templateResolver, $pluginManager);

$viewModel = new ViewModel(/** Stuff **/);

$content = $renderer->renderSomething('template', $viewModel);

// Much like Mezzio…
```

The existing `render($nameOrModel, $vars = null)` could be preserved with additional methods such as `renderModel(ViewModel $model): string` - where the view model is reponsible for exposing the template to render and `renderTemplate(string $name, iterable $variables)`

### Removals

- `Renderer\FeedRenderer` and `Strategy\FeedStrategy` can go in the bin. If you have a feed, what do you need a templating engine for? Surely `laminas-feed` has your back there?
- `Renderer\JsonRenderer` and `Strategy\JsonStrategy` are surely dead? It is much simpler to return a Json encoded Response from Middleware/Controller that to muck about in the view layer…
- Console rendering is already deprecated for removal in 3.0 - #126
- Anything that mentions a request or a response.

### The renderer…

…itself should be vastly simplified, removing all property and method overloading and any non-essential state. Runtime mutation such as setting the template resolver, filter chain etc should be removed. 

### Filtering content post render

Can be satisfied by `laminas-filter` as it is now, or made implementation agnostic by providing a way to register callbacks. 

### MVC

Obviously, this will all be thoroughly incompatible with MVC, but 2.x isn't going anywhere.

### Events

May become moot - in MVC for example, given a renderer it can populate its response body with the output on its own terms.

### View Helpers

[...]

PR #237, opened 2024-01-25
Deprecate the JSON View Helper for Removal in v3.0

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### DIFF
--- a/docs/book/v2/helpers/json.md
+++ b/docs/book/v2/helpers/json.md
@@ -1,5 +1,11 @@
 # Json
 
+> WARNING: **Deprecated**
+>
+> The JSON view helper has been deprecated and will be removed in version 3.0.
+> There is no replacement;
+> however, it is trivial to encode data by using PHP's built-in [`json_encode`](https://www.php.net/json_encode) function.  
+
 When creating views that return JSON, it's important to also set the appropriate
 response header.  The JSON view helper does exactly that. In addition, by
 default, it disables layouts (if currently enabled), as layouts generally aren't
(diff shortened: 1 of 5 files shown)


Ocramius (approved):

docs/book/v2/migration/preparing-for-v3.md
```diff
@@ -44,3 +44,13 @@ This deprecation can be safely ignored but in order to prepare for its removal i
...
 
 The impact of this future removal will affect templates that use a regular short open tag for general PHP code, i.e. `<? $i = 1; echo $i ?>` in environments where `short_open_tag` is **off**. To mitigate the impact of this removal, you should ensure that, where relevant, all of your templates use the full `<?php` open tag. Use of the short echo tag `<?=` is unaffected.
+
+## View Helper Changes
+
+### [Json View Helper](../helpers/json.md)
+
+In version 3, the JSON view helper will no longer set response headers _(For MVC requests)_ when used.
+If you are using this feature, you will need to refactor your controllers to return the correct mime-type rather than relying on the view helper to do it for you.
```
froschdesign: Can we add a code example here?

froschdesign: Is this a BC break for a laminas-mvc based application?

gsteel: I might need some help here as it's been a very long time since I used MVC. I'm guessing something like this would suffice:

```php
class MyController extends AbstractActionController {
    public function someAction(): void
    {
        // ...
        $this->response->setHeader('Content-Type', 'application/json');
        // ...
    }
}
```

I don't believe this will be a BC break for MVC. AFAICT, MVC uses the `JSONRenderingStrategy` - I couldn't find any references to this view helper in the MVC codebase. Perhaps @Xerkus can confirm.

froschdesign:
```php
final class AlbumController extends Laminas\Mvc\Controller\AbstractActionController
{
    public function listAction(): array
    {
        // …

        $this->getResponse()->getHeaders()->addHeaderLine(
            'Content-Type',
            'application/json'
        );

        return [
            'albums' => ['…'],
        ];
    }
}
```

Xerkus: MVC controllers do not handle content type response headers (users might do whatever, so not a given). It is done by view strategies.  Json strategy sets content type for this. Those are being extracted into laminas-mvc-view I believe.

ext-json was moved to php core extensions with php 8 and always available. This helper is obsolete imo. 

I do not know what use is there left for this helper. May be set default json_encode flags? I would suggest deprecating it and drop for v3 instead.

Via laminas-json there is a special handling of objects with `toJson()` or `toArray()` methods. It is not documented or mentioned here. Expressions are mentioned in docs but declared deprecated.


Also, this was introduced in zf1 long long time ago: https://github.com/zendframework/zf1/blob/release-1.12.0/library/Zend/View/Helper/Json.php#L85

Xerkus: Actually, json response strategy is used with json view model. This helper is used with default strategy when array needs to be prepared by .phtml before getting emitted. A hack, really, because json strategy does not give controller specific hook points to prepare view model data for json format.

A mess any way you look at it.

froschdesign: I think @Xerkus is correct and we should remove this helper. I see no reason why a view script should be used if a JSON output is to be created.
And the idea was to remove the `JsonStrategy` and `FeedStrategy` with their related renderers. (see #149)
But so far we have no user-friendly replacement like the [`JsonResponse` or `XmlResponse` of laminas-diactoros](https://docs.laminas.dev/laminas-diactoros/v3/custom-responses/).

Xerkus (approved):

docs/book/v2/helpers/json.md
```diff
@@ -1,5 +1,11 @@
 # Json
 
+> WARNING: **Deprecated**
+>
+> The JSON view helper has been deprecated and will be removed in version 3.0.
+> There is no replacement;
+> however, it is trivial to encode data by using PHP's built-in [`json_encode`](https://www.php.net/json_encode) function.
```
froschdesign: This is enough to create the desired output:

```suggestion
WARNING: **Deprecated**
The JSON view helper has been deprecated and will be removed in version 3.0.
There is no replacement; however, it is trivial to encode data by using PHP's built-in [`json_encode`](https://www.php.net/json_encode) function.  
```

https://oprypin.github.io/markdown-callouts/index.html#usage

gsteel: Thanks @froschdesign

PR #238, opened 2024-01-25
[v3] Remove the Json view helper and update the migration guide

gsteel:


docs/book/v3/helpers/json.md
```diff
@@ -1,21 +1,16 @@
...
-Content-Type: application/json
+```php
+<?= $this->json(['example' => 'payload']) ?>
 ```
 
-Most XmlHttpRequest libraries look for this header when parsing responses to
-determine how to handle the content.
-
-## Basic Usage
+The helper accepts a second argument `$jsonOptions`, an associative array with one possible key: `prettyPrint`.
```
froschdesign: If there is only one possible key, why is an array needed here?

gsteel: I know. Originally I made this a `bool` but re-considered. If anyone is using this helper, it'll be annoying to break BC for aesthetics. Additionally, it occurred to me that users might want to escape output for HTML, at that point, we can add an `escape` key or any others. That said, I'd personally prefer a `bool` but I didn't think it was worth the BC break. I'll happily change it if we all aggree it's stupid to require an obtuse array for a single flag! I'm personally in favour of the bool.

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -105,3 +93,8 @@ function gravatarImage(
     string $rating = 'g'
 );
 ```
+
+#### Json
+
+The deprecated Json view helper has been removed.
+To encode data to Json for output in a view, you can call [`json_encode`](https://www.php.net/json_encode) directly.
```
Xerkus, line 100: Should this mention previous behavior of using `toJson()` and `toArray()` when encoded data is an object?
https://github.com/laminas/laminas-json/blob/53ff787b20b77197f38680c737e8dfffa846b85b/src/Json.php#L99-L107

gsteel: Done!

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -98,3 +93,11 @@ function gravatarImage(
...
+
+If you were relying on behaviour that was previously available via `laminas-json`, for example, calling object methods `toArray` or `toJson` prior to encoding, you should make the relevant objects implement `JsonSerializable`.
+You can find documentation on the `JsonSerializable` interface [on the PHP website](https://www.php.net/manual/en/class.jsonserializable.php).
```
froschdesign: Links to the official PHP documentation should omit the language part of the URL, as the PHP documentation is available in multiple languages and redirects to the correct language automatically by using the `Accept-Language` header of the user's browser.

```suggestion
You can find documentation on the `JsonSerializable` interface [on the PHP website](https://www.php.net/manual/class.jsonserializable.php).
```

https://github.com/php/web-php/blob/c89bec8faddf40e281664555ea835857b1d1c64b/include/langchooser.inc#L5-L8

gsteel:
Merging this as-is without further review. All issues with docs have currently been addressed.

Issue #239, opened 2024-01-25
Too strict parameter type for `ViewModel#__construct` `$variables`


boesing:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | [2.33.0](https://github.com/laminas/laminas-view/releases/tag/2.33.0)

#### Summary

The `ViewModel` now only allows maps but its inherited from `JsonModel` which does properly render list values as of now.

#### Current behavior

```
Argument 1 of Laminas\View\Model\JsonModel::__construct expects ArrayAccess<string, mixed>|iterable<string, mixed>|null, but list<array<array-key, mixed>>
```

#### How to reproduce

```php
new Laminas\View\Model\JsonModel([1,2,3]);
```

#### Expected behavior

No error. Since a list is a valid JSON type, lists should be allowed.

RuesimOfCode:
I just wanted to create the same issue :)


I have one point to add.

Maybe there are more cases to match all forms of valid json. `json_encode` uses just `mixed` as first argument.

Example:

```
new JsonModel([1, 'test', null, true])
```

results in valid json

```
[
    1,
    "test",
    null,
    true
]
```

when returned by a HTTP endpoint

Best regards

gsteel:
`JsonRenderer` and `JsonModel` will be gone in 3.0 making `int` keyed view variables a moot point. 3.0 goes all in on `<string, mixed>`

PR #245, opened 2024-03-18
added missing allowed types to constructor (#239)

ppaulis:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This bugfix addresses the issue #239 by allowing lists in the constructor of `ViewModel`.

### DIFF
--- a/src/Model/ViewModel.php
+++ b/src/Model/ViewModel.php
@@ -76,7 +76,7 @@ class ViewModel implements ModelInterface, ClearableModelInterface, RetrievableC
     /**
      * Constructor
      *
-     * @param  null|array<string, mixed>|Traversable<string, mixed>|ArrayAccess<string, mixed> $variables
+     * @param  null|array<string, mixed>|Traversable<string, mixed>|ArrayAccess<string, mixed>|array<int, mixed>|Traversable<int, mixed>|ArrayAccess<int, mixed> $variables
      * @param  null|array<string, mixed>|Traversable<string, mixed> $options
      */
     public function __construct($variables = null, $options = null)
--- a/src/Model/ViewModel.php
+++ b/src/Model/ViewModel.php
@@ -76,8 +76,9 @@ class ViewModel implements ModelInterface, ClearableModelInterface, RetrievableC
     /**
      * Constructor
      *
-     * @param  null|array<string, mixed>|Traversable<string, mixed>|ArrayAccess<string, mixed>|array<int, mixed>|Traversable<int, mixed>|ArrayAccess<int, mixed> $variables
-     * @param  null|array<string, mixed>|Traversable<string, mixed> $options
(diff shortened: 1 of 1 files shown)


src/Model/ViewModel.php
```diff
@@ -76,7 +76,7 @@ class ViewModel implements ModelInterface, ClearableModelInterface, RetrievableC
     /**
      * Constructor
      *
-     * @param  null|array<string, mixed>|Traversable<string, mixed>|ArrayAccess<string, mixed> $variables
+     * @param  null|array<string, mixed>|Traversable<string, mixed>|ArrayAccess<string, mixed>|array<int, mixed>|Traversable<int, mixed>|ArrayAccess<int, mixed> $variables
```
Ocramius: I'd say that all the `int`-keyed values should be rejected.

Yes, it works in practice: no, it shouldn't be allowed, and there's no guarantee that it will.

Ocramius (changes requested):
Closing as invalid: these types are too wide, even if the more internal `setVariables()` is not as restrictive

src/Model/ViewModel.php
ppaulis: @Ocramius Thanks for the quick reply :-)
So, to be sure that I understand you correctly : If I want to return in my Laminas Api Tools RPC Controller something like :
```
[
  {
    "field1": "value1",
    "field2": "value2"
  },
  {
    ..
  },
  ..
]
```
JsonModel/ViewModel shouldn't be used?

Ocramius: ViewModel: probably not.
JsonModel: yes, should be OK 🤔

ppaulis: @Ocramius ok, do you have a suggestion of how to implement this?
I see a few possibilities here, but ultimately it leads me to thinking that `JsonModel` shouldn't inherit from `ViewModel` and that `ModelInterface` contains a lot of stuff that is not necessary for `JsonModel`...

Ocramius: @ppaulis `JsonModel` could declare its own `__construct`, with more types (`mixed` comes to mind, to be honest)

ppaulis: @Ocramius so we continue relying on the same code with new tests for JsonModel specifically (haven't deep-dived into the existing tests yet), testing the discussed JSON format. And when the day comes that the ViewModel would no longer support integer-indexes, we would need to think about how to keep the JsonModel running.

if that's ok, I'll send you over a new PR until end of the week.

froschdesign: @gsteel 
What are you planning for version 3.0 on this?

gsteel: Rendering Strategies are, IMO, MVC specific - it's been a while since I had time to allocate to view, but I think my initial plan was to move the rendering strategies to the as yet un-released `laminas-mvc-view` component.

PR #248, opened 2024-06-04
Fix layout() return type

### LINKED ISSUE #247: PhpRenderer @method for layout() return type is incomplete
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.35.x, 3.0.x

#### Summary

`PhpRenderer` `@method` for method `layout()` is incomplete. The method actually returns `ModelInterface|Layout`, cf. `Layout::__invoke`. Actually the annotation there is wrong, too, since it never returns `null`.

https://github.com/laminas/laminas-view/blob/3.0.x/src/Renderer/PhpRenderer.php#L72
https://github.com/laminas/laminas-view/blob/3.0.x/src/Helper/Layout.php#L38

https://github.com/laminas/laminas-view/blob/2.35.x/src/Renderer/PhpRenderer.php#L76
https://github.com/laminas/laminas-view/blob/2.35.x/src/Helper/Layout.php#L38

#### Current behavior

Static analysis of code calling `PhpRenderer::layout()` and passing the result to `PhpRenderer::render()` fails

#### How to reproduce

```
<?php
$renderer = new \Laminas\View\Renderer\PhpRenderer();
$renderer->render($renderer->layout());
```

#### Expected behavior

Static analysis reports no error with the example code

InvisibleSmiley:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Bring `PhpRenderer` `@method` for `layout()` in sync with `Layout::__invoke` and remove `null` from that method's return type union (`@return`). Result reflects reality and thus helps static analysis tools to reduce detection of false negatives.

See #247.

### DIFF
--- a/src/Helper/Layout.php
+++ b/src/Helper/Layout.php
@@ -35,7 +35,7 @@ class Layout extends AbstractHelper
      * Otherwise, attempts to set the template for that view model.
      *
      * @param null|string $template
-     * @return Model|null|self
+     * @return Model|self
      */
     public function __invoke($template = null)
     {
--- a/src/Renderer/PhpRenderer.php
+++ b/src/Renderer/PhpRenderer.php
@@ -45,7 +45,7 @@ use function sprintf;
  * mark them as part of the internal implementation, and thus prevent conflict
  * with variables injected into the renderer.
  *
- * Convenience methods for build in helpers (@see __call):
+ * Convenience methods for built-in helpers (@see __call):
  *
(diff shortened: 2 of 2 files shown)


gsteel (approved):
Thanks @InvisibleSmiley 👍

PR #258, opened 2024-11-22
Deprecate navigation view helpers for removal in 3.0

gsteel:
Related to https://github.com/laminas/laminas-navigation/pull/66

PR #259, opened 2024-11-25
Remove Navigation View Helpers

gsteel:
|   Q           |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes

### Description

- Deprecated in https://github.com/laminas/laminas-view/pull/258
- Added to the navigation component in https://github.com/laminas/laminas-navigation/pull/66


docs/book/v3/migration/v2-to-v3.md
```diff
@@ -101,3 +101,12 @@ To encode data to Json for output in a view, you can call [`json_encode`](https:
...
 If you were relying on behaviour that was previously available via `laminas-json`, for example, calling object methods `toArray` or `toJson` prior to encoding, you should make the relevant objects implement `JsonSerializable`.
 You can find documentation on the `JsonSerializable` interface [on the PHP website](https://www.php.net/manual/class.jsonserializable.php).
+
+#### Navigation
+
+The deprecated navigation view helpers such as `Breadcrumbs`, and `Menu` etc have been removed and can now be found in [the `laminas-navigation` component](https://docs.laminas.dev/laminas-navigation/helpers/intro/).
+
+As such, the namespace for these helpers has changed from `Laminas\View\Navigation` to `Laminas\Navigation\View\Helper`, so if you have referenced the FQCNs of these helpers in your code, you will need to update them accordingly.
+
+At the time of writing, the `laminas-navigation` 2.x series of releases is not compatible with `laminas-view` 3.x and a major release of the `laminas-navigation` component is not available.
```
froschdesign: Would it help or be easier if we moved the helpers to a separate package? The original plans included the support of different template systems, laminas-view would be one of them.

gsteel: I'm not sure it'll be _easier_, but it would make evolution of all 3 libs easier in the long-run.

The problem here is the appetite to maintain them. As they stand, there is a lot that needs improvement and refactoring with the old helpers, so, assuming we move them to a new lib instead, will it be acceptable to have a v1 release that is largely the same as they are now, warts and all, or is there an expectation they'd be polished up a bit for a view v2 compatible release?

Additionally, what would the lib be called and what would the NS be? `Laminas\NavigationView`?

froschdesign: > …but it would make evolution of all 3 libs easier in the long-run.

That was the idea.

> …will it be acceptable to have a v1 release that is largely the same as they are now…

That would probably be the quickest option.

> …or is there an expectation they'd be polished up a bit for a view v2 compatible release?

You mean version 3 of laminas-view?

> Additionally, what would the lib be called and what would the NS be? `Laminas\NavigationView`?

The original idea was that the HTML would be created by the laminas-navigation component and then there would be simple integrations in different template systems. laminas-view is one of them and then Plates, Twig, etc. – like we have in Mezzio.
So it could be `laminas-navigation-laminasview`, `laminas-navigation-twig`, …

gsteel: > > …or is there an expectation they'd be polished up a bit for a view v2 compatible release?
> 
> You mean version 3 of laminas-view?

I meant here, that we'd likely want to release v1 of nav-view to be compatible with view v2 so there is an easy upgrade path - it will need a major release to support SMv4, so nav-view v2 would be compatible with view v3 and SMv4

Ocramius:
Need #266 first

gsteel:
Rebased on updated 3.0.x

Ocramius (approved):
Thanks @gsteel!

Ocramius:
@gsteel BTW, post-merge, this led to a psalm issue:

```
test/Strategy/JsonStrategyTest.php:330:26: UnusedMethodCall: The call to Laminas\Http\Response::getContent is not used (see https://psalm.dev/209)
```

gsteel:
Thanks, I've fixed that in another patch I think.

PR #268, opened 2025-04-29
Refactor `HeadTitle` Helper, Remove MVC and i18n Dependencies

### LINKED ISSUE #252: `laminas-translator` support
|    Q        |   A
|------------ | ------
| laminas-view | 2.35.0
| laminas-i18n | 2.28.0
| laminas-translator | 1.0.0

#### Summary

`laminas-i18n`'s `\Laminas\I18n\Translator\TranslatorInterface` is marked deprecated, and the comment says it's going to be replaced by `laminas/laminas-translator`'s `Laminas\Translator\TranslatorInterface`

I think `\Laminas\I18n\Translator\TranslatorAwareTrait` should start using `laminas-translator`'s Interface instead.

gsteel:
- Removes inheritance from `HeadTitle` and replaces the relevant magic methods with concrete methods, preserving most existing runtime mutation.
- Removes various getters from `HeadTitle`
- Introduces `StatefulHelperInterface` that can and will be applied to other helpers so that it can be used to programmatically reset any state built up for each request cycle
- Swaps dependencies `mvc`, `i18n` and `mvc-i18n` with `laminas-translator`.
- Removes the translator initializer from the plugin manager, i.e. If your helper requires a translator, then write a factory for it.
- Introduces a factory for the `HeadTitle` helper with a well-defined set of constructor options
- Deletes the `TranslatorAwareTrait`


composer.json
```diff
@@ -34,6 +34,7 @@
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.10.1",
+        "laminas/laminas-translator": "^1.1",
```
Ocramius, line 37: This swaps MVC (`-dev` dependency) for a non-dev dependency: is that really a good idea? 🤔

gsteel: Translator is a single interface pkg just for the translation contract. I think it's reasonable to import that symbol considering there's no equivalent PSR 🤷‍♂️

froschdesign (approved):
Nice improvements! 👍🏻

gsteel:
@Ocramius do you object to the new dependency? or can I merge this as-is?

composer.json
Ocramius: Given this, I think we're fine merging :+1:

Ocramius (approved):

PR #272, opened 2025-04-29
Upgrade Psalm to v6, Mark shipped classes as soft `@final` to signal change to `final` keyword in 3.0

gsteel:
Also upgrades Laminas CS to 3.x

### DIFF
--- a/src/Helper/FlashMessenger.php
+++ b/src/Helper/FlashMessenger.php
@@ -22,6 +22,8 @@ use function sprintf;
  *
  * @deprecated This helper will be removed in version 3.0 of this component.
  *     At that time, it will be available in laminas/laminas-mvc-plugin-flashmessenger.
+ *
+ * @final
  */
 class FlashMessenger extends AbstractHelper
 {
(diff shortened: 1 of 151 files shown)


src/Helper/HtmlQuicktime.php
```diff
@@ -6,7 +6,11 @@
 
 use function array_merge;
 
-/** @deprecated */
+/**
+ * @deprecated
```
froschdesign: The description is missing.

src/Helper/FlashMessenger.php
```diff
@@ -22,6 +22,8 @@
  *
  * @deprecated This helper will be removed in version 3.0 of this component.
  *     At that time, it will be available in laminas/laminas-mvc-plugin-flashmessenger.
+ *
+ * @final
```
froschdesign, line 26: Is unnecessary because the class is obsolete, but okay.

froschdesign (changes requested):

froschdesign (approved):
Go for it!

PR #277, opened 2025-04-30
Refactor `Layout`, `HtmlAttributes`, `HtmlList` and `BasePath` helpers

gsteel:
Also adds a factory for the plugin manager.

Generally speaking, the helpers have…

- Deprecations removed
- Inheritance removed
- Types added
- Final keyword added


src/ConfigProvider.php
```diff
@@ -15,11 +15,29 @@ final class ConfigProvider
     public function __invoke(): array
     {
         return [
-            'dependencies'       => $this->getDependencies(),
+            'dependencies' => $this->getDependencies(),
+            /**
+             * The top-level configuration key for defining custom view helpers.
+             *
+             * This option should use the `ServiceManagerConfiguration` array format
```
Ocramius, line 37: Should we type against it then?

src/ConfigProvider.php
```diff
@@ -15,11 +15,29 @@ final class ConfigProvider
...
+             */
+            'view_helpers'       => [],
             'view_manager'       => [
+                /**
+                 * Encoding passed to the Escaper and possibly used in other view-related configuration
+                 */
```
froschdesign: laminas-mvc does not set the encoding via the view manager. Can we move this to `view_helper_config`?

src/ConfigProvider.php
```diff
@@ -15,11 +15,29 @@ final class ConfigProvider
...
                 'encoding' => 'utf-8',
+
+                /**
+                 * The base path is provided to the BasePath view helper. This is the historic location for this
+                 * configuration item defined in MVC apps.
+                 */
+                'base_path' => null,
```
froschdesign: laminas-mvc cannot use the new major version of laminas-view, so the question arises if we need the configuration key `view_manager` at all.
Related to https://github.com/laminas/technical-steering-committee/pull/223

gsteel: I've moved some stuff to `view_helper_config` anyway, and retained searching in `view_manager` for fewer BC issues.

#281 refactors template resolvers, so we'll be writing factories soon to set those up. MVC puts all of this under `view_manager`, Mezzio under `templates` (Both at the top-level).

I'm starting to wonder if we could do with a top-level key that is specific to this library such as `laminas-view` where we nest everything else…

src/Helper/HtmlAttributes.php
```diff
@@ -9,18 +9,14 @@
...
  */
-class HtmlAttributes extends AbstractHelper
+final class HtmlAttributes
 {
-    use DeprecatedAbstractHelperHierarchyTrait;
-
-    private Escaper $escaper;
+    private readonly Escaper $escaper;
 
     public function __construct(?Escaper $escaper = null)
```
Ocramius: BTW, since these helpers are always built via DIC, perhaps make the ctor argument mandatory, and remove instantiation?

It's just an idea / something I noticed

src/Helper/Service/BasePathFactory.php
```diff
@@ -4,32 +4,27 @@
...
 use Laminas\View\Helper\BasePath;
 use Psr\Container\ContainerInterface;
 
 use function assert;
-use function is_array;
 use function is_string;
 
 final class BasePathFactory
 {
+    /** @psalm-suppress MixedAssignment */
```
Ocramius: Eh, let's suppress per-statement, not just everywhere? Mixed assignments are very sneaky 

Fine to dismiss this review comment, if you believe it's not as important.

src/ConfigProvider.php
```diff
@@ -15,14 +15,29 @@ final class ConfigProvider
     public function __invoke(): array
     {
         return [
-            'dependencies'       => $this->getDependencies(),
-            'view_manager'       => [
-                'encoding' => 'utf-8',
-            ],
+            'dependencies' => $this->getDependencies(),
+            /**
```
Ocramius, line 34: Since we are introducing them now, can we document the type of additional keys in this array at type level?

Ocramius (approved):

src/ConfigProvider.php
gsteel: Yes! I also sealed the shape 👍

src/Helper/HtmlAttributes.php
gsteel: Good point. If Escaper's constructor ever changes, however unlikely, this ensures deps are fully delegated 👍

src/Helper/Service/BasePathFactory.php
gsteel: You're right of course! I've imported the "View Config" array shape and forced that var on `$container->get('config')`

Ocramius (approved):
:shipit:

Issue #279, opened 2025-04-30
[RFC]: Removal of Paginator Dependency


gsteel:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 3.0.0  |
| BC Break?           | Yes |

### Goal

Get ServiceManager v4 installed

### Background

We have 1 helper related to Paginator: `PaginationControl` - ideally we will remove it because Paginator is not and won't be ready for SMv4 on any reasonable time frame

### Considerations

Should we put this view helper into the paginator lib, or, create a whole new lib just for this one helper?

### Proposal(s)

None really…

gsteel:
Tagging @froschdesign and the @laminas/technical-steering-committee for guidance

froschdesign:
The same question should be asked about the `Identity` helper.

gsteel:
> The same question should be asked about the `Identity` helper.

Yes, absolutely, but it doesn't currently block SMv4 support.

gsteel:
Linking https://github.com/laminas/technical-steering-committee/pull/222 for discussion at the next TSC meet

gsteel:
Closed via #293

PR #280, opened 2025-04-30
Upgrade Laminas ServiceManager to v4

### LINKED ISSUE #263: Update to laminas-servicemanager v4
The `composer.json` dependency list includes `laminas-servicemanager` at version `^3.21.0`:
https://github.com/laminas/laminas-view/blob/2e441988c93d11b05856dc9e54f96369efa4e5b5/composer.json#L35

In March 2024 `laminas-servicemanager` 4.0.0 has been released (see https://github.com/laminas/laminas-servicemanager/releases/tag/4.0.0), current latest version is 4.4.0. So could you please update to `laminas-servicemanager` v4.

gsteel:
<del>This patch is cut from #277, so there's a load of stuff that's not relevant</del>, however, it shows what's required to get SMv4 installed:

- Drop dependency on Paginator and remove the pagination control view helper
- Upgrade `filter` to v3 (This might need to become a hard dependency)
- Using a filter chain with `PhpRenderer` will take some effort. It's worth considering allowing an arbitrary callable instead, i.e. `callable(string):string` can be attached to the renderer for filtering prior to output. Laminas filters could solve this, and a generic callable would be more flexible for users, easy to test and we'd be able to drop a soft/hard dependency too.

This patch is also food for #279


psalm-baseline.xml
```diff
@@ -695,39 +652,35 @@
...
+      <code><![CDATA[$this->initializers]]></code>
+      <code><![CDATA[$this->initializers]]></code>
+      <code><![CDATA[$this->initializers]]></code>
+      <code><![CDATA[$this->initializers]]></code>
+    </DeprecatedProperty>
     <DocblockTypeContradiction>
       <code><![CDATA[! $events]]></code>
       <code><![CDATA[! $events]]></code>
     </DocblockTypeContradiction>
+    <MethodSignatureMismatch>
```
Ocramius, line 510: Can this mismatch be fixed?

Ocramius (commented):
Looks okay - the `ViewModel` helper is a cosmic horror though: it should probably not have existed 😱

psalm-baseline.xml
gsteel: No, well, maybe. Probably removing the PSR\Container stub will work at the expense of inference there. Also, overriding `get` with repeated doc blocks in the plugin manager will work - it was like that before.

gsteel:
This patch is now good to go with the exception of the soft dep on laminas-filter which is pinned to an unstable branch.

Once filter v3 is released, we can review and merge this 👍

composer.json
```diff
@@ -46,7 +46,7 @@
         "laminas/laminas-authentication": "^2.18",
         "laminas/laminas-coding-standard": "~3.0.1",
         "laminas/laminas-feed": "^2.23",
-        "laminas/laminas-filter": "^2.40",
+        "laminas/laminas-filter": "^3.0.x-dev",
```
Ocramius: As mentioned by @gsteel, need this to be `^3.0`

Ocramius (approved):

composer.json
gsteel: Thanks @Ocramius - There's a vote going on in the TSC channel on slack to release filter 👍

PR #281, opened 2025-05-01
Refactor Template Resolvers

gsteel:
Amends the `ResolverInterface` to `resolve(non-empty-string): non-empty-string` with failure conditions meaning an exception is thrown.

Resolvers are largely intact but they all now correctly implement the interface and expected behaviour.

TODO:

- [x] Decide how we will expect template maps and directory lists to be configured
- [x] Add factories for the resolvers and wire them up in ConfigProvider
- [x] Decide what fetching `ResolverInterface` from the container will provide by default
- [x] Refactor documentation for v3 with less MVC emphasis

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -21,7 +21,15 @@ use Laminas\View\Helper\Service\EscaperFactory;
  *     view_manager?: array{
  *         base_path?: non-empty-string|null,
  *         encoding?: non-empty-string,
- *     }
+ *         template_map?: array<string, string>,
+ *         template_path_stack?: list<string>,
+ *         prefix_template_path_stack?: array<string, string>,
+ *         default_template_suffix?: non-empty-string,
+ *     },
+ *     templates?: array{
+ *         extension?: non-empty-string,
+ *         map?: array<string, string>,
+ *     },
  * }
  */
 final class ConfigProvider
@@ -54,6 +62,45 @@ final class ConfigProvider
                  */
                 'base_path' => null,
             ],
+            'view_manager'       => [
+                /**
+                 * Templates configured here will be provided to the TemplateMapResolver.
+                 * This is conventional for an MVC app
+                 */
+                'template_map' => [
+                    // 'template-name' => 'path/to/template.phtml',
+                ],
+                /**
+                 * Templates configured here will be provided to the TemplatePathStack resolver.
+                 * This is conventional for an MVC app
+                 */
+                'template_path_stack' => [
+                    // 'path/to/a/directory/of/templates',
+                ],
+                /**
+                 * Templates configured here will be provided to the PrefixPathStackResolver
+                 */
+                'prefix_template_path_stack' => [
+                    // 'prefix' => 'path/to/a/directory/of/templates',
+                ],
+                /**
+                 * The default template suffix is configured here in MVC applications
+                 */
+                //'default_template_suffix' => 'phtml',
+            ],
+            'templates'          => [
+                /**
+                 * Templates configured here will be provided to the TemplateMapResolver
+                 * This is conventional for a Mezzio app
+                 */
+                'map' => [
+                    // 'template-name' => 'path/to/template.phtml',
+                ],
+                /**
+                 * The default template filename extension is configured here. The default is 'phtml'
+                 */
+                //'extension' => 'phtml',
+            ],
         ];
     }
 
@@ -62,12 +109,17 @@ final class ConfigProvider
     {
         return [
             'factories' => [
-                HelperPluginManager::class  => HelperPluginManagerFactory::class,
-                Escaper::class              => EscaperFactory::class,
-                Renderer\PhpRenderer::class => Renderer\PhpRendererFactory::class,
+                HelperPluginManager::class              => HelperPluginManagerFactory::class,
+                Escaper::class                          => EscaperFactory::class,
+                Renderer\PhpRenderer::class             => Renderer\PhpRendererFactory::class,
+                Resolver\AggregateResolver::class       => Resolver\Factory\AggregateResolverFactory::class,
+                Resolver\PrefixPathStackResolver::class => Resolver\Factory\PrefixPathStackResolverFactory::class,
+                Resolver\TemplateMapResolver::class     => Resolver\Factory\TemplateMapResolverFactory::class,
+                Resolver\TemplatePathStack::class       => Resolver\Factory\TemplatePathStackResolverFactory::class,
             ],
             'aliases'   => [
                 Renderer\RendererInterface::class => Renderer\PhpRenderer::class,
+                Resolver\ResolverInterface::class => Resolver\AggregateResolver::class,
             ],
         ];
     }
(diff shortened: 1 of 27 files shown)


gsteel:
This is blocked by #277. The plugin manager must be available in the container in order to test the aggregate resolver factory

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
                  */
                 'base_path' => null,
             ],
+            'view_manager'       => [
+                /**
+                 * Templates configured here will be provided to the TemplateMapResolver.
+                 * This is conventional for an MVC app
+                 */
+                'template_map' => [
+                    // 'template-name' => 'path/to/template.phtml',
```
Ocramius: Let's use `__DIR__` in the example code

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
...
+                ],
+                /**
+                 * Templates configured here will be provided to the TemplatePathStack resolver.
+                 * This is conventional for an MVC app
+                 */
+                'template_path_stack' => [
+                    // 'path/to/a/directory/of/templates',
```
Ocramius: Let's use `__DIR__` in the example code

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
...
+                ],
+                /**
+                 * Templates configured here will be provided to the PrefixPathStackResolver
+                 */
+                'prefix_template_path_stack' => [
+                    // 'prefix' => 'path/to/a/directory/of/templates',
```
Ocramius: Let's use `__DIR__` in the example code

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
...
+                ],
+                /**
+                 * The default template suffix is configured here in MVC applications
+                 */
+                //'default_template_suffix' => 'phtml',
```
Ocramius, line 89: Why is this commented out? We can provide our default :thinking:

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
...
+            ],
+            'templates'          => [
+                /**
+                 * Templates configured here will be provided to the TemplateMapResolver
+                 * This is conventional for a Mezzio app
+                 */
+                'map' => [
+                    // 'template-name' => 'path/to/template.phtml',
```
Ocramius: Let's use `__DIR__` in the example code

src/ConfigProvider.php
```diff
@@ -54,6 +62,45 @@ public function __invoke(): array
...
+                ],
+                /**
+                 * The default template filename extension is configured here. The default is 'phtml'
+                 */
+                //'extension' => 'phtml',
```
Ocramius, line 102: Why is this commented out? We can provide our default 🤔

src/Helper/Service/Configuration.php
```diff
@@ -51,4 +53,22 @@ public static function viewEncoding(
...
+
+    /**
+     * Retrieve the default template suffix from conventional locations
+     *
+     * @param non-empty-string $fallbackDefault
+     * @return non-empty-string
+     */
+    public static function defaultTemplateSuffix(
+        ContainerInterface $container,
+        string $fallbackDefault = self::DEFAULT_TEMPLATE_SUFFIX,
```
Ocramius: this parameter is never set by anyone: let's kill it

src/Resolver/AggregateResolver.php
```diff
@@ -72,77 +39,27 @@ public function count()
...
-             */
-            $resource = $resolver->resolve($name, $renderer);
-            if (is_string($resource)) {
-                // Resource found; return it
-                $this->lastSuccessfulResolver = $resolver;
-                return $resource;
+            try {
+                return $resolver->resolve($name);
+            } catch (TemplateCannotBeFound) {
+                continue;
```
Ocramius: Note: using exceptions for this can be really expensive, and this is critical path for most apps.

Ocramius (changes requested):
Only skimmed this: I think the exception used for flow control is something we should avoid here.

A `TemplatePathStack` resolver (used by default by most apps) will really suffer here, IMO.

src/Resolver/AggregateResolver.php
gsteel: 👍 I've added, used and tested `ResolverInterface::has($name): bool`

src/ConfigProvider.php
gsteel: Because this is potentially configured in `view_manager.default_template_suffix` and `templates.extension`, specifying the default in `templates.extension` will make the former impossible to configure which will be confusing for users.

I thought it better to not _configure_ it at all and rely on a fallback constant for when it's completely undefined

src/Resolver/AggregateResolver.php
```diff
@@ -53,13 +53,24 @@ public function attach(ResolverInterface $resolver, int $priority = 1): self
     public function resolve(string $name): string
     {
         foreach ($this->queue as $resolver) {
-            try {
-                return $resolver->resolve($name);
-            } catch (TemplateCannotBeFound) {
+            if (! $resolver->has($name)) {
```
Ocramius: This is now worse, because we do double the I/O 🤣

Sorry!

gsteel: I/O is only happening in the PathStack variant.

What do you think is a good solution here?

Internal in-memory cache? Or changing `Resolver::get(string): string|false`?

Ocramius: > I/O is only happening in the PathStack variant.

Which is the most common one, heh

I think sticking to `|false` is the least evil change right now

Ocramius (approved):
Only managed to skim the patch, as I'm on the run - a second review by someone else may be a good idea

src/Resolver/TemplateMapResolver.php
```diff
@@ -174,21 +147,15 @@ public function get($name)
...
-     * Resolve a template/pattern name to a resource the renderer can consume
-     *
-     * @param string $name
-     * @return false|string
-     */
-    public function resolve($name, ?Renderer $renderer = null)
+    public function resolve(string $name): string|false
     {
-        return $this->get($name);
+        return $this->map[$name] ?? false;
```
froschdesign, line 159: The method `get` uses `has` and throws an exception, but is `false` the desired result here?

gsteel: Yes, all implementations of `resolve` have the signature `resolve(non-empty-string): non-empty-string|false`.

I initially returned only string or throw, but Marco pointed out the performance impact of throwing on a hot path so reverted to `string|false`

Both `has` and `get` are existing methods on this class - I would have removed them but couldn't find a convincing reason to break even more BC apart from immutability for the sake of it.

froschdesign: > but Marco pointed out the performance impact of throwing on a hot path

Good point! 👍🏻

froschdesign (approved):

PR #282, opened 2025-05-02
Remove the deprecated `DeclareVars` Helper, Introduce factory for `PhpRenderer`

gsteel:
`DeclareVars` should probably go in the bin. I've certainly never used it. It is probably used in the wild though - Perhaps we should also deprecate it now for 4.0?

This patch introduces a factory for the `PhpRenderer`, which will get updated soon to include constructor deps of the renderer, hence the reason for _not_ just mapping it to `InvokableFactory`.


src/Helper/DeclareVars.php
```diff
@@ -45,39 +42,39 @@ class DeclareVars extends AbstractHelper
...
 
     /**
      * Set a view variable
      *
-     * Checks to see if a $key is set in the view object; if not, sets it to $value.
-     *
-     * @param  string $key
-     * @param  string $value Defaults to an empty string
-     * @return void
+     * Checks to see if a $key is set in the view object; if not, sets it to $value with a default of an empty string.
```
Ocramius: WTF? I know this is pre-existing, but how is this supposed to be useful in any way?! 

Let's instead kill the helper for 3.0.x

froschdesign: The helper is not documented and does not really seem to be helpful. Remove it!

Ocramius (approved):
🔥

PR #284, opened 2025-05-04
Refactor `RenderChildModel`

gsteel:
- Makes internal deps visible with constructor injection
- Adds factory for the helper
- Adds the final keyword
- Removes inheritance
- Removes `RenderChildModel::render()` in favour of `__invoke`


Ocramius (approved):
Nice cleanup! :+1:

PR #288, opened 2025-05-05
Refactor `Partial` and `PartialLoop` helpers

gsteel:
Makes dependencies visible in the constructor and implements factories to satisfy them.

`Partial` Changes

- Adds types
- Added `final` keyword
- Removes inheritance
- Inaccessible or Dropped Methods
  - `getView`
  - `setView`
- Implements `StatefulHelperInterface`
- Deprecates usage of non-iterable objects representing the variables passed to the template

`PartialLoop` Changes

- Adds types
- Added `final` keyword
- Removes inheritance
- Inaccessible or Dropped Methods
  - `getView`
  - `setView`
  - `loop`
- Implements `StatefulHelperInterface`
- Deprecates usage of non-iterable objects representing the variables passed to the template

---

"View Variables" are still a big source of type safety problems, I've glossed over these for another patch by expanding the baseline with a few issues.

---

I've deprecated the duck-typing for objects that implement `toArray` with a hard deprecation using `trigger_error`.

Potentially, we could add this to 2.x and instead remove it for 3.0


Ocramius (approved):
I am OK with this patch, but having never used `PartialLoop` myself, I'm a bit unclear on whether this is fine as-is.

gsteel:
> I am OK with this patch, but having never used `PartialLoop` myself, I'm a bit unclear on whether this is fine as-is.

I may come back and have another refactor of the `PartialLoop` helper in future - I've never used it either, but existing behaviour has been preserved as far as the tests are concerned, however, it should really be accepting `list<iterable<string, mixed>>` for its view variables and I can't improve that until the renderer itself is a lot less squishy

PR #295, opened 2025-05-06
[RFC] Drop the `View` class, alternative view models and rendering strategies.

gsteel:
As broadly MVC related functionality, these classes should be moved to the `laminas-mvc-view` package.

Enables the removal of `feed` and `http` dependencies

This simplifies view to basic template rendering on the premise that:

JSON - Why do we need a templating lib for producing JSON output? I just cannot see the use case for this apart from supporting legacy apps and bad decisions
<del>XML/RSS Feeds - Surely this is the job of dedicated tooling. In any app MVC or other, we should encourage users to use a service such as laminas-feed to create the XML document and emit the XML in a response with the correct headers etc. In MVC this could be done via a controller plugin, and its so trivial in Mezzio…</del> _(Feed integration is gone in #316)_

There will still need to be some kind of top level class to implement 2-step view in way that makes sense. For example, it should probably implement a way of configuring a layout, aggregating default params/variables and rendering composed output to a string etc.

<del>In future patches, I intend to remove the `initialisers` in the helper plugin manager - this will enable dropping the dependency on event-manager too.</del> _(Initializers are gone in #311)_


froschdesign:
We do not need a render strategy for feeds here, this should be solved elsewhere – I like [the custom response of laminas-diactoros](https://docs.laminas.dev/laminas-diactoros/v3/custom-responses/). (Reference: https://github.com/laminas/laminas-view/issues/149#issuecomment-1049756487)

But does this also mean that the strategic concept is being removed completely?

gsteel:
I struggle to see the use-case for using different renderers during the same rendering cycle, so yes, I suggest removing Rendering Strategies completely.

If you have some kind of XML based layout thing going on, you've already written a custom renderer that isn't shipped here, so it's not a stretch to fill in the gaps yourself, compose the necessary deps and then own all of that customisation.

`laminas-feed` could ship Middleware/Handlers for rendering and returning feeds just by including the relevant PSR deps.

froschdesign:
> As broadly MVC related functionality, these classes should be moved to the `laminas-mvc-view` package.

Is this still relevant? I think not. 😄 

> I struggle to see the use-case for using different renderers during the same rendering cycle, so yes, I suggest removing Rendering Strategies completely.

Would have been a good topic for yesterday's meeting.

> `laminas-feed` could ship Middleware/Handlers for rendering and returning feeds just by including the relevant PSR deps.

That is the idea. 👍🏻

src/Strategy/PhpRendererStrategy.php
```diff
@@ -1,113 +0,0 @@
...
-
-/** @final */
-class PhpRendererStrategy extends AbstractListenerAggregate
-{
-    /**
-     * Placeholders that may hold content
-     *
-     * @var array<array-key, string>
-     */
-    protected $contentPlaceholders = ['article', 'content'];
```
froschdesign, line 21: Will there be an alternative for the use with these placeholders?

gsteel: I'm not sure why there are 'conventional' placeholders. Why wouldn't you, as a user, just specify the placeholders that you care about in your templates?

This strategy has a hard dependency on a specific view helper which is fragile, because the user could easily override the placeholder (Fixable via DI though).

Assigning the output from child templates to a variable in the parent template is also a job for the view model no? Or whatever orchestrates the 2-step view.

I would imagine that `View` will eventually be replaced with something that consumes the renderer and the plugin manager along with necessary settings for a 2-step view - this is where the responsibility for rendering the requested content to the layout model would occur. It would also be responsible for resetting the state of stateful view helpers for each rendering cycle.

gsteel:
@laminas/technical-steering-committee 

This is currently blocking further work on view v3 for me - Once merged, I'll be able to move forward with refactoring `PhpRenderer`, and re-implementing `View` as an entry point for the 2-step render.

I also have plans to introduce an `@internal` class that will be used for isolating the scope _inside_ `PhpRenderer` that will be responsible for the `include($templateFile)`.

Getting rid of this legacy code will mean less work fixing broken tests etc, so I'd appreciate reviews 🙏

froschdesign (approved):
Go for it! 👍🏻

PR #298, opened 2025-05-07
Refactor `HeadLink` Helper

gsteel:
- Removes inheritance entirely
- Adds `final` keyword
- Makes hidden dependencies visible as constructor arguments
- Removes all magic
- Conditional stylesheets are no longer supported - This is _really_ outdated practice. CSS, and browsers have come a long way.
- Remove documented magic methods for `Next`, `Prev`, and `Alternate`, i.e. `appendNext` etc. The most prevalent use case is for stylesheets, beyond that, where do we stop? Why not `appendPreload` or `appendManifest` etc
- Abandons validation of link attributes because it gets out of date quite quickly and its also more complex than the previous implementation implies.
- Setting items at specific array indexes is no longer supported. Given the main use-case is aggregating multiple style sheets from different templates, setting order via numeric indexes is very brittle. It's more likely that users would append from content and then prepend in the layout (for stylesheets at least).
- `HeadLink` is no longer iterable or countable or array-like, and you can't fetch a container to capture output buffers


gsteel:
Needs #297 for tests to pass on lowest

froschdesign (approved):

gsteel:
Sorry @froschdesign - Could you please re-review this now I've added the docs? Thanks ever so much :)

docs/book/v3/helpers/head-link.md
```diff
@@ -1,68 +1,117 @@
...
+
+## Doctype Aware
+
+The `HeadLink` helper composes the [`Doctype` view helper](doctype.md) internally to alter the output for XML based doctypes.
+When you have configured the doctype to `XHTML 1.1` for example, the closing tag of the rendered links will become a self-closing `<link />` as opposed to `<link>`.
+
+## Further Reading
+
+- [Mozilla's MDN Web Docs on the `<link>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link)
+- [MDN Docs on the `rel` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel)
```
froschdesign:
```suggestion
- [Mozilla's MDN Web Docs on the `<link>` element](https://developer.mozilla.org/docs/Web/HTML/Reference/Elements/link)
- [MDN Docs on the `rel` attribute](https://developer.mozilla.org/docs/Web/HTML/Reference/Attributes/rel)
```

froschdesign:
@gsteel 
Great improvement! 👍🏻

gsteel:
Thanks @froschdesign

froschdesign (approved):

docs/book/v3/helpers/head-link.md
```diff
@@ -1,68 +1,117 @@
 # HeadLink
 
-The HTML `<link>` element is increasingly used for linking a variety of
-resources for your site: stylesheets, feeds, favicons, trackbacks, and more. The
-`HeadLink` helper provides a simple interface for creating and aggregating these
-elements for later retrieval and output in your layout script.
+The [HTML `<link>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link) is used for linking a variety of resources for your site: stylesheets, preloads, feeds, favicons, trackbacks, and more.
```
gsteel:
```suggestion
The [HTML `<link>` element](https://developer.mozilla.org/docs/Web/HTML/Reference/Elements/link) is used for linking a variety of resources for your site: stylesheets, preloads, feeds, favicons, trackbacks, and more.
```

Issue #18, opened 2019-12-31
Invokable view variable shouldn't be invoked


weierophinney:
Hi,

when passing an invokable object to the view and trying to access it, the object is being invoked automatically which may not be intended.

```
// Controller action
public function testAction()
{
    $viewModel = new ViewModel();
    $viewModel->test = new class {
        public function __invoke(array $params) {}
        public function getName() {
            return 'test';
        }
    };
    return $viewModel;
}

// View
echo $this->test->getName();
```

**Expected output:**
`test`

**Actual output:**
`Argument 1 passed to class@anonymous::__invoke() must be of the type array, none given`

The reason is here (`Zend/View/Variables.php`):

```
public function offsetGet($key)
{
    if (!$this->offsetExists($key)) {
        if ($this->isStrict()) {
            trigger_error(sprintf(
                'View variable "%s" does not exist',
                $key
            ), E_USER_NOTICE);
        }
        return;
    }

    $return = parent::offsetGet($key);

    // If we have a closure/functor, invoke it, and return its return value
    if (is_object($return) && is_callable($return)) {
        $return = call_user_func($return);
    }

    return $return;
}
```

The process of passing and accessing variables should not take care of the kind of the variables passed.

I really don't get the point of invoking them automatically. If someone needs automatic invocation we have view helpers. My suggestion is to remove the automatic invocation or if you can't let go, add an interface check.

---

Originally posted by @antiphp at https://github.com/zendframework/zend-view/issues/85

weierophinney:
Interestingly, that's a new feature, and it was done to allow defining helpers without requiring implementation of `HelperInterface` (see #60). As such, any such change at this point would be a BC break. We'll have to evaluate if this should be the behavior going forward, or something we want to revert and/or change with the next major version of zend-view.


---

Originally posted by @weierophinney at https://github.com/zendframework/zend-view/issues/85#issuecomment-241492956

weierophinney:
But how am I supposed to pass invokable objects then?

In short to show that my situation is not special (at least IMHO): I want to pass a command-like object to the view (ReplaceTranslation) and I want to show a preview of that command output (via its invokable method) as well as its name. I have more commands.

Workarounds I was thinking about:
- pass just the return values of all required methods. Disadvantage: Usage of arrays again instead of objects
- dump the object into an array-structure and pass it (optionally rebuild the object with that array-structure in the view again). Disadvantage: a) Arrays again b) Overhead c) Some object can't be rebuild easily
[...]

PR #302, opened 2025-05-09
[RFC] Invokable View Variables are (should not be) Automatically Invoked

gsteel:
This PR is, for now, just a reminder that this behaviour should be either:

- Kept and documented
- Removed, or
- Made optional via configuration

I personally favour option 2 - Removing the behaviour so that invokable variables passed to the view model are just variables, therefore requiring explicit invokation.

Removal will not affect view helpers registered with the plugin manager.

Issue #18 brought this to my attention - it was behaviour I was unaware of.

CC @laminas/technical-steering-committee

### DIFF
--- a/test/PhpRendererTest.php
+++ b/test/PhpRendererTest.php
@@ -35,6 +35,7 @@ use function realpath;
 use function restore_error_handler;
 use function set_error_handler;
 use function str_replace;
+use function trim;
 
 use const E_WARNING;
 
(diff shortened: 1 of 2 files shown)


gsteel:
Started a poll about this here: https://github.com/laminas/laminas-view/discussions/304

gsteel:
Closed in favour of #351

PR #305, opened 2025-05-12
Refactor `HeadMeta` view helper

gsteel:
- Removes inheritance
- Introduce `final` keyword
- Remove all "Magic", replacing common calls with concrete methods
- Conditionals are no longer supported
- Validation of meta tags has been removed. Instead of querying the doctype and throwing exceptions when `<meta property>` is added to a non-RDFa document, any meta tag is simply allowed, with possibly invalid attributes. The validation rules are much bigger and more numerous than this library can hope to cover.
- Makes hidden dependencies visible as constructor arguments
- `HeadMeta` is no longer iterable or countable or array-like, and you can't fetch a container to capture output buffers
- Adding meta tags at specific offsets is no longer possible

### DIFF
--- a/docs/book/v3/helpers/head-meta.md
+++ b/docs/book/v3/helpers/head-meta.md
@@ -1,82 +1,55 @@
 # HeadMeta
 
-The HTML `<meta>` element is used to provide meta information about your HTML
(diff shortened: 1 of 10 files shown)


froschdesign (approved):
LGTM 👍🏻

PR #309, opened 2025-05-14
Deprecate the `HtmlPage` view helper

gsteel:
I wonder if anyone has used this view helper within the last 10 years or so?

### DIFF
--- a/src/Helper/HtmlPage.php
+++ b/src/Helper/HtmlPage.php
@@ -6,7 +6,12 @@ namespace Laminas\View\Helper;
 
 use function array_merge;
 
-/** @final */
+/**
+ * @deprecated Since 2.40.0 - This helper is deprecated for removal in 3.0 - It's main purpose as a helper to avoid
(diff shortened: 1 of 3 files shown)


samsonasik (approved):

src/Helper/HtmlPage.php
```diff
@@ -6,7 +6,12 @@
 
 use function array_merge;
 
-/** @final */
+/**
+ * @deprecated Since 2.40.0 - This helper is deprecated for removal in 3.0 - It's main purpose as a helper to avoid
```
froschdesign, line 10: Please add also the notice to the documentation:

- https://docs.laminas.dev/laminas-view/v2/helpers/html-object/#htmlpage-helper
- https://github.com/laminas/laminas-view/blob/c300bb23dd8cccc0c337b655018a7ff234801720/docs/book/v2/helpers/html-object.md?plain=1#L15

froschdesign (changes requested):

froschdesign (approved):

PR #311, opened 2025-05-14
Remove Initializers from the `HelperPluginManager`

gsteel:
Helpers should use dependency injection via the constructor rather than relying on magic setter injection at runtime.

Also, the `ServiceManager::$initializers` property is deprecated


froschdesign (approved):
Much cleaner! 👍🏻

PR #314, opened 2025-05-14
Deprecate `AbstractHtmlElement` Base Class

gsteel:
All helpers will/have been refactored to remove this class from their inheritance hierarchy. It will no longer be necessary in 3.0.

It also has dubious utility for users and authors of custom helpers


Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #316, opened 2025-05-15
Remove `Laminas\Feed` related code

gsteel:
The FeedStrategy, renderer and model classes should be part of a _(to be created)_ `laminas-feed` integration library


froschdesign (approved):
Go for it! 👍🏻

PR #317, opened 2025-05-15
Deprecate Laminas\Feed related code

gsteel:
See #316


gsteel:
Merging this without review because #316 is already reviewed and merged

PR #318, opened 2025-05-16
The Big Placeholder Refactor

gsteel:
This PR is quite big because the placeholder/container/contraption is/was entangled in a lot of stuff…

- HTML Attribute handling has been refined and improved in c7f3ad79c388c149d2d6a2e68c8ac6195af6d1f1
- There is still a helper `Container` but it is now internal and none of the helpers expose it via their individual apis
- Refactors `HeadStyle` helper, removing inheritance and cleaning up with DI
- Refactors `HeadScript` and `InlineScript` doing the same kind of things. These 2 both inherit from an abstract class because they have identical behaviour. It could have been reduced to a single class with DI aliases, but it didn't seem worth breaking BC in that way. The abstract has been shored up to prevent users abusing inheritance.
- A lot of helpers now have similar constructors, so a 'GenericFactory' has been introduced to reduce the number of shipped factories
- The `Placeholder` helper has been refactored without inheritance etc

TODO:

- [x] The new `GenericFactory` doesn't yet have a test
- [x] Update docs for `HeadStyle`
- [x] Update docs for `HeadScript` and `InlineScript`
- [x] Update docs for `Placeholder`
- [x] Update migration guide


froschdesign:
Great improvement! 👍🏻

docs/book/v3/helpers/head-script.md
```diff
@@ -1,112 +1,37 @@
...
-> These include `id`, `charset`, `crossorigin`, `defer`, `integrity`,
-> `language`, `src`, and `type`. However, some JavaScript frameworks, notably
-> [Dojo](http://www.dojotoolkit.org/), utilize custom attributes in order to
-> modify behavior. To allow such attributes, you can enable them via the
-> `setAllowArbitraryAttributes()` method:
->
-> ```php
-> $this->headScript()->setAllowArbitraryAttributes(true);
-> ```
+# HeadScript & InlineScript
```
froschdesign: Can we rename the file to something like `head-inline-script` and add some redirects? The filename should nearly the same like the first headline.

froschdesign (approved):
Great! 👍🏻 
(I will go through the entire documentation before the release.)

gsteel:
Thanks @froschdesign 👍

PR #320, opened 2025-05-20
Refactor `Cycle` View Helper

gsteel:
Stops implementing `Iterator` - not only was it unnecessary, but the `next` and `prev` methods _must_ return `self` in order to fulfill the `Stringable` behaviour contract which conflicts with the return types on `Iterator`.

Drops useless methods `getName` and `valid`

Marks as final and adds types.


froschdesign (approved):
LGTM 👍🏻

PR #322, opened 2025-05-22
Remove `ServerUrl`, `AbstractHelper` and empty `HelperInterface`

gsteel:
With the removal of `ServerUrl`, there are no longer any helpers that extend from `AbstractHelper` so it is also removed.

With the removal of `AbstractHelper`, we can also remove the `getView` and `setView` methods from `HelperInterface`. The initializers were removed in #311 so there are no callers of `setView` any more.

The reason for the removal of `ServerUrl` rather than refactoring it, is that frameworks should implement this helper themselves. MVC should probably re-implement what was here and Mezzio already has a replacement helper.

`HelperInterface` is now empty, however it is worth keeping it as a marker interface for the `HelperPluginManager` - effectively it allows users to register helpers that are not invokable, but prevent registration of any arbitrary non-callable object without intentionally marking the object as a view helper.


froschdesign (approved):
Deleting code is always good! 👍🏻

gsteel:
> Deleting code is always good! 👍🏻
👍 

😀 - I've just got some docs changes to push so I'll ping you for another review if that's ok

froschdesign (approved):
Very nice! 👍🏻

gsteel:
Thanks @froschdesign

PR #326, opened 2025-05-22
Swap `FilterChain` for a `callable`

gsteel:
Renames `setFilterChain` to `setFilter` in `PhpRenderer`, changing argument type to a well-defined callable shape.

`FilterChain` is callable, therefore it could still stand in here without the coupling to `Laminas\Filter`

Removes the `getFilter` method because "getters".

There are no changes to docs here yet - mainly because PhpRenderer needs a refactor and the filter will likely become a constructor dependency amongst other things

### DIFF
--- a/src/Renderer/PhpRenderer.php
+++ b/src/Renderer/PhpRenderer.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Laminas\View\Renderer;
 
 use ArrayAccess;
-use Laminas\Filter\FilterChain;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\View\Exception;
 use Laminas\View\Helper\Doctype;
@@ -81,7 +80,7 @@ class PhpRenderer implements Renderer, TreeRendererInterface
     /**
      * @var string Rendered content
      */
-    private $__content = '';
+    private string $__content = '';
 
     /**
      * @var bool Whether to render trees of view models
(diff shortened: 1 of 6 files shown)


src/Renderer/PhpRenderer.php
```diff
@@ -479,6 +471,7 @@ public function render($nameOrModel, $values = null)
                 ob_start();
                 $includeReturn   = include $this->__file;
                 $this->__content = ob_get_clean();
+                assert($this->__content !== false);
```
Ocramius: Can't happen

Ocramius (changes requested):
One little change needed, but otherwise massive improvement

src/Renderer/PhpRenderer.php
froschdesign: @Ocramius 
Sure?
> Returns the contents of the active output buffer on success or [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure.

https://www.php.net/manual/function.ob-get-clean.php

gsteel: This was just for Psalm because of the documented `string|false` return type. I'll revert and expand the baseline 👍

froschdesign (approved):

src/Renderer/PhpRenderer.php
gsteel: With the `string` property type, we'd get a `TypeError` before the `false` could proliferate

PR #329, opened 2025-05-29
Refactor `PartialLoop` Test

gsteel:
Organises the used templates in a separate directory so they are easier to identify.

Removes duplicated and unnecessary tests.


gsteel:
Merging without review - All behaviour is still covered and line coverage is 100%

PR #330, opened 2025-05-29
Introduce `HelperPluginManagerInterface`

gsteel:
In order to have a simple way to reset all state that has built up in view helpers / plugins, the plugin manager overrides `Container::get` to populate a list of stateful plugins that have been used during the rendering cycle.

Also re-writes "Advanced Usage", de-emphasising MVC related docs.

### DIFF
--- a/docs/book/v3/helpers/advanced-usage.md
+++ b/docs/book/v3/helpers/advanced-usage.md
@@ -2,7 +2,7 @@
 
 ## Writing & Registering Custom Helpers
 
-`laminas-view` provides a "Plugin Manager" implementation that provides access to what we call "View Helpers", typically, from within a template.
+`laminas-view` provides a "Plugin Manager" implementation that provides access to what we call "View Helpers" from within a template.
 
 "View Helpers", or "View Plugins" must be invokable objects or closures.
 
@@ -186,7 +186,8 @@ final class TrivialHelper implements StatefulHelperInterface
         return '<p>Some HTML</p>';
     }
     
-    public function resetState() : void{
+    public function resetState() : void
+    {
         $this->count = 0;
     }
 }
--- a/docs/book/v3/helpers/advanced-usage.md
+++ b/docs/book/v3/helpers/advanced-usage.md
@@ -166,7 +166,20 @@ This state can become problematic when your app is running on long-lived process
 
 In the case of the placeholder view helper, if this state was not reset for each request, the data stored inside that helper would keep growing and growing affecting the data output on consecutive renders for different users of your website or application.
 
-`laminas-view` ships an interface for view helpers to implement so that state can be automatically reset at the end of each rendering cycle: `Laminas\View\Helper\StatefulHelperInterface`.
+`laminas-view` ships an interface for view helpers to implement so that state can be automatically reset at the end of each rendering cycle: `Laminas\View\Helper\StatefulHelperInterface`:
+
+```php
+namespace Laminas\View\Helper;
+
+/**
+ * This interface defines a view helper that maintains state and provides a way to reset that state
+ */
+interface StatefulHelperInterface
+{
+    public function resetState(): void;
+}
+
+```
 
 Here's a trivial example implementation:
 
@@ -204,7 +217,7 @@ In order to register these, you'd need to either return them from a factory, or 
 return [
     'view_helpers' => [
         'services' => [
-            'someHelper' => static fn(string $value): string => strrev($value),
+            'reverse' => static fn(string $value): string => strrev($value),
         ],
     ],
 ];
--- a/docs/book/v3/helpers/advanced-usage.md
+++ b/docs/book/v3/helpers/advanced-usage.md
@@ -1,276 +1,212 @@
-# Advanced usage of helpers
+# Advanced Usage of Helpers
 
-## Registering Helpers
+## Writing & Registering Custom Helpers
 
-`Laminas\View\Renderer\PhpRenderer` composes a *plugin manager* for managing
-helpers, specifically an instance of `Laminas\View\HelperPluginManager`, which
-extends `Laminas\ServiceManager\AbstractPluginManager`, which is itself an
-extension of `Laminas\ServiceManager\ServiceManager`.  `HelperPluginManager` is a
-specialized service manager, so you can register a helper/plugin like any other
-service (see the [Service Manager documentation](https://docs.laminas.dev/laminas-servicemanager/configuring-the-service-manager/)
-for more information).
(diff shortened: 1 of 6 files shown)


src/HelperPluginManagerInterface.php
```diff
@@ -0,0 +1,17 @@
...
+use Laminas\ServiceManager\PluginManagerInterface;
+use Laminas\View\Helper\HelperInterface;
+
+/** @extends PluginManagerInterface<HelperInterface|callable> */
+interface HelperPluginManagerInterface extends PluginManagerInterface
+{
+    /**
+     * Resets the internal state built up in any view helpers so that further rendering cycles are not polluted
+     */
+    public function resetState(): void;
```
Ocramius, line 16: This is a very welcome addition :+1:

Ocramius (approved):
From my PoV, this is a very welcome addition. Would still want other opinions before proceeding with this design, though: in other environments I worked with, the entire container was being rebuilt (which is also a fine approach)

gsteel:
My plan here is hopefully to drop "Rendering Strategies" and the `ViewEvent` stuff leaving just `Laminas\View\View` - This class will be repurposed so that it's the main entry point for rendering templates and it will be configured with things like which layout template to use, strict variables etc, the theory being that each call to `View::render()` will be a "complete" render, and internally, all state will be reset each time it is called.

Plugin Managers don't provide any way to reset/clear built services AFAIK, nor extract config, so that would need to implemented over there first if tearing down the container was the preferred route.

docs/book/v3/helpers/advanced-usage.md
```diff
@@ -1,276 +1,212 @@
...
+## Writing & Registering Custom Helpers
 
-`Laminas\View\Renderer\PhpRenderer` composes a *plugin manager* for managing
-helpers, specifically an instance of `Laminas\View\HelperPluginManager`, which
-extends `Laminas\ServiceManager\AbstractPluginManager`, which is itself an
-extension of `Laminas\ServiceManager\ServiceManager`.  `HelperPluginManager` is a
-specialized service manager, so you can register a helper/plugin like any other
-service (see the [Service Manager documentation](https://docs.laminas.dev/laminas-servicemanager/configuring-the-service-manager/)
-for more information).
+`laminas-view` provides a "Plugin Manager" implementation that provides access to what we call "View Helpers", typically, from within a template.
```
weierophinney:
```suggestion
`laminas-view` provides a "Plugin Manager" implementation that provides access to what we call "View Helpers" from within a template.
```

docs/book/v3/helpers/advanced-usage.md
```diff
@@ -1,276 +1,212 @@
...
+## Dealing with State Buildup
+
+Sometimes it is necessary to aggregate state inside view helpers to achieve certain functionality.
+The shipped [`Placeholder` view helper](../helpers/placeholder.md) is a good example of this: Aggregating values in templates for output in a parent template, or layout for example.
+
+This state can become problematic when your app is running on long-lived processes such as [RoadRunner](https://roadrunner.dev/), [Swoole](https://www.php.net/manual/book.swoole.php), or [FrankenPHP](https://frankenphp.dev/) amongst others, because the helper stays in memory for multiple requests rather than being destroyed after each successful request.
+
+In the case of the placeholder view helper, if this state was not reset for each request, the data stored inside that helper would keep growing and growing affecting the data output on consecutive renders for different users of your website or application.
+
+`laminas-view` ships an interface for view helpers to implement so that state can be automatically reset at the end of each rendering cycle: `Laminas\View\Helper\StatefulHelperInterface`.
```
weierophinney: I think it would be good to show the interface signature here as well, so that readers know what they need to implement.

docs/book/v3/helpers/advanced-usage.md
```diff
@@ -1,276 +1,212 @@
...
     {
         $this->count++;
-        $output  = sprintf("I have seen 'The Jerk' %d time(s).", $this->count);
-        $escaper = $this->getView()->plugin('escapehtml');
-        return $escaper($output);
+        
+        return '<p>Some HTML</p>';
+    }
+    
+    public function resetState() : void{
```
weierophinney:
```suggestion
    public function resetState() : void
    {
```

docs/book/v3/helpers/advanced-usage.md
```diff
@@ -1,276 +1,212 @@
...
-$helper = new MyModule\View\Helper\LowerCase;
-// ...do some configuration or dependency injection...
+```php
+// config/autoload/helpers.global.php
 
-$view->getHelperPluginManager()->setService('lowercase', $helper);
+return [
+    'view_helpers' => [
+        'services' => [
+            'someHelper' => static fn(string $value): string => strrev($value),
```
weierophinney: Maybe use the name "reverse" here to tie it back to the orignal example earlier in the document.

weierophinney (changes requested):
No real code changes, but some doc changes. 

But this was a primary reason why I moved away from laminas-view some time ago, so this will definitely be welcome!

src/HelperPluginManager.php
```diff
@@ -174,4 +183,29 @@ public function validate(mixed $instance): void
...
+     * @template InstanceParam of HelperInterface
+     * @param class-string<InstanceParam>|string $id Service name of plugin to retrieve.
+     * @return ($id is class-string<InstanceParam> ? InstanceParam : InstanceType)
+     */
+    public function get(string $id): mixed
+    {
+        /** @psalm-var InstanceType $plugin Unfortunately this type needs forcing */
+        $plugin = parent::get($id);
+        if ($plugin instanceof StatefulHelperInterface) {
+            $this->statefulHelpers[spl_object_hash($plugin)] = $plugin;
```
samsonasik: `spl_object_id()` is exists since php 7.2, I think we can use `spl_object_id()`.

Refs:

- https://github.com/php/php-src/pull/2611
- https://developer.wordpress.com/2023/08/24/speedier-php-execution-in-wordpress-6-3/

**Benchmark**

**Before** **_call in loop took: 13.639ms_**

https://3v4l.org/4rUL4

**After**  **_call in loop took: 4.835ms_**

https://3v4l.org/jKtO6

gsteel: That's a pretty big difference 😃

gsteel:
Anyone have any objections if I merge this? Looks like all feedback is resolved

froschdesign (approved):
LGTM 👍🏻

froschdesign:
(I will recheck the documentation before the final release.)

PR #331, opened 2025-05-29
Introduce IDE Helper Interface

gsteel:
Instead of adding magic `@method` docs to the PhpRenderer, ship an interface so users can more easily combine it with their own and declare in view scripts `/** @var TemplateInterface $this */` for decent autocompletion in their IDE.

Note - expanded baseline is for some old tests calling magic methods on the renderer - these will be addressed later.


src/Renderer/PhpRenderer.php
```diff
@@ -42,36 +40,6 @@
...
- * @method mixed|null identity()
- * @method \Laminas\View\Helper\InlineScript inlineScript()
- * @method Model|\Laminas\View\Helper\Layout layout(string|null $template = null)
- * @method string|\Laminas\View\Helper\Partial partial(string|Model|null $name = null, iterable|object|null $values = null)
- * @method string|\Laminas\View\Helper\PartialLoop partialLoop(string|null $name = null, iterable|object $values = [])
- * @method \Laminas\View\Helper\Placeholder placeholder(string|null $placeholder = null)
- * @method string renderChildModel(string $child)
- * @method void renderToPlaceholder(string|Model $script, string $placeholder)
- * @method \Laminas\View\Helper\ViewModel viewModel()
- * @method string gravatarImage(string $emailAddress, int $imageSize = 80, array $imageAttributes = [], string $defaultImage = 'mm', string $rating = 'g')
```
Ocramius: I don't think we should drop them from here: perhaps we can `@mixin TemplateInterface` to preserve BC?

src/TemplateInterface.php
```diff
@@ -0,0 +1,304 @@
...
+use Laminas\View\Helper\Placeholder\Position;
+use Laminas\View\Helper\RenderChildModel;
+use Laminas\View\Helper\ViewModel;
+use Laminas\View\Model\ModelInterface;
+use Stringable;
+
+/**
+ * This interface is used only for the purposes of auto-completion in your IDE
+ *
+ * You should not implement this interface, and as such, it will not retain backwards compatibility
```
Ocramius, line 36: This is probably going to backfire spectacularly, if somebody doesn't read this :D

Ocramius: I wonder if marking this `@final` could help?

gsteel: The whole idea is that you create a `MyProjectTemplate extends LaminasTemplate, LaminasFormTemplate, OtherInterface` so that in your templates you just declare `/** @var MyProjectTemplate $this */` and have _all_ the view helpers working nicely.

My plan is that this also can form the basis of static analysis of phtml files - I already do this in my projects so we might be able to ship a psalm plugin that uses this, or possibly custom interfaces for the purposes of SA.

src/Renderer/PhpRenderer.php
gsteel: I'm planning on killing `__call` on `PhpRenderer` along with all the aggregation of view variables

gsteel:
@Ocramius I added a `TemplateAnalyzer` - maybe you have some pointers? I'm certainly no expert with Psalm internals 😬 - it does "work" well enough though

gsteel:
Reverted addition of the template analyzer for Psalm to keep this patch to the interface only

froschdesign (approved):

PR #334, opened 2025-06-12
Remove `ViewModel` "Options"

gsteel:
The view model does not need options as part of its contract. The only feasible option might be 'template', but this should be a property of the view model rather than an option.

Users can create their own view models based on `ModelInterface` if options are necessary for some reason.


froschdesign (approved):
The options were never documented and I do not see any use case.
Go for it! 👍🏻

PR #337, opened 2025-06-23
Remove `TreeRendererInterface`

gsteel:
There is now only 1 renderer - `canRenderTrees()` and `setCanRenderTrees()` are unused in the library except for tests to check the flag can be toggled.

There is no longer a use-case for this flag and requirements for nested view models can be extracted from the view models themselves


froschdesign (approved):
LGTM 👍🏻

PR #338, opened 2025-06-23
Partial Refactor of `Variables` with BC Breaking Changes to `RendererInterface`

gsteel:
The roles of `Variables`, `ViewModel` and `PhpRenderer` overlap considerably.

For the time being, `Variables` has been refactored to have much better types, though it will likely be dropped - the `ViewModel` is already a mixed property bag, and a separate type just for this job seems unnecessary, but I haven't removed it yet.

Similarly `ViewModel` and `ModelInterface` are now mixed property bags for `<string, mixed>`.

Further refactoring of `ViewModel` will also be required.

The headline changes for the 2 variable bags are dropping `ArrayAccess`, requiring `Strict Variables` settings up-front and emitting `UndefinedVariableException` when the obvious occurs.

`RendererInterface` no longer has the pretty useless `getEngine` method.

Types and docs for `render` have been tightened up


src/Renderer/PhpRenderer.php
```diff
@@ -358,20 +337,20 @@ public function render($nameOrModel, $values = null)
...
-        $this->addTemplate($nameOrModel);
-        unset($nameOrModel); // remove $name from local scope
+        $this->addTemplate($templateNameOrModel);
+        unset($templateNameOrModel); // remove $name from local scope
 
         $this->__varsCache[] = $this->vars();
 
-        if (null !== $values) {
-            $this->setVars($values);
+        if (null !== $variables) {
```
froschdesign: Are the Yoda conditions still used?! I think not.

gsteel: Yeah - I hate yoda conditions too!

froschdesign (approved):
Nice clean up! 👍🏻

PR #340, opened 2025-06-24
Refactor template rendering

gsteel:
Introduces the `Template` class to encapsulate all the necessary magic and exposure of the assigned variables.

Reduces `PhpRenderer` to a single `render` method with runtime mutation of the optional filter, making the class pretty much a stateless service.

Introduces `RenderingFailedException` with a number of clear error messages covering all possible templating related outcomes.

Reworks tests and test templates for more obvious definition of behaviour.

`PhpRenderer` no longer has _any_ magic and no longer collects view variables.


froschdesign (approved):
The `Template` class is quite simple, I like it! 👍

PR #343, opened 2025-06-26
Add Sarif report to GitHub actions

gsteel:


gsteel:
Merging this without review as an experiment to see if this is worth doing on more libs

PR #346, opened 2025-06-26
Set up mutation testing with `roave-infection-static-analysis-plugin`

gsteel:
Configuration is a bit convoluted, but we need PHP 8.2 minimum to run with `roave-infection-static-analysis-plugin` - installation is impossible on 8.1 combined with Psalm v6


froschdesign (approved):
LGTM 👍

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #348, opened 2025-07-16
Re-introduce `Laminas\View\View`

gsteel:
The 'View Manager' is configured with a default layout template so that it is convenient for users to call `render` and automatically render a template as a 'child model' of a layout.

Also…

- Adds native types to Model interface and implementation
- Re-organises some factories and internal configuration details

TODO

- Docs will need a complete re-work in view of recent MVC decisions
- `ModelInterface` needs further refactoring - for example, making it an iterable of child models is unnecessary. If it should be iterable, better to iterate over variables, i.e. `ModelInterface<string, mixed>`. Additional simplification will be useful so that it is trivial to write implementations instead of extending from `ViewModel`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,9 +46,9 @@
     },
     "require-dev": {
         "laminas/laminas-coding-standard": "~3.1.0",
-        "phpunit/phpunit": "^10.5.47",
+        "phpunit/phpunit": "^10.5.48",
         "psalm/plugin-phpunit": "^0.19.5",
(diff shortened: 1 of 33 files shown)


gsteel:
I'd appreciate a review from @laminas/technical-steering-committee to check the direction of this 🙏

froschdesign (approved):
The new `View` class is super clear! 👍🏻

PR #356, opened 2025-08-01
Add Rector to Tooling

gsteel:
Adds rector as a standalone tool with a CI action that will fail when rector detects code improvements.


froschdesign (approved):
> …will fail when rector detects code improvements.

Nice! 👍🏻

Makefile
```diff
@@ -87,3 +89,11 @@ composer-require-checker: ## Check for symbols from un-declared dependencies
 
 qa: composer-validate cs sa test composer-require-checker docs-lint ## Run all QA Checks
 .PHONY: qa
+
+rector: ## Run Rector and show the diff
+	tools/rector/vendor/bin/rector process --dry-run -vv -c tools/rector/rector.php
```
samsonasik, line 94: FYI, rector has github formatter, see PR:

- https://github.com/rectorphp/rector-src/pull/6747

to test, `--output-format=github` may can be used.

gsteel: Thanks @samsonasik - I'll add that in now 👍

PR #357, opened 2025-08-07
HelperPluginManager: drop template requirement as the class can be used to retrieve all sort of HelperInterface

Slamdunk:
Hello, I'm following the new good deprecation warnings, but this one is wrong IMHO:

```
Method My\View\Helper\Foobar::__construct() has parameter $helperPluginManager with
    generic class Laminas\View\HelperPluginManager but does not specify its types: InstanceType
```

As far as I understand the `@template InstanceType of [...]` annotation on both `PluginManager` and `HelperPluginManager` make no sense because it's their very nature to retrieve all sort of _helpers_ as long as they are a `HelperInterface` type, and this is already enforced by `@extends AbstractPluginManager<HelperInterface|callable>`


gsteel (approved):
Thanks @Slamdunk 👍

PR #363, opened 2025-09-05
Replace `templatemap_generator.php` with a `Symfony\Console` command

### LINKED ISSUE #218: Port the Template Map Generator to a Laminas CLI command
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | yes _(will replace `bin/templatemap_generator.php`)_

#### Summary

Replace the current `templatemap_generator.php` script with a command that can be run from `laminas-cli`

Suggested in https://github.com/laminas/laminas-view/pull/217#discussion_r1301368554

gsteel:
Closes #218


Xerkus:
Package does not require symfony console. Symfony had breaking changes incompatible with existing commands just recently. Imo this should be in a satellite package with explicit dependencies.

gsteel:
> Package does not require symfony console. Symfony had breaking changes incompatible with existing commands just recently. Imo this should be in a satellite package with explicit dependencies

I don't disagree.

I also hate having symbols used in `src` that are not dependencies in `require`.

There's very little point going to all the trouble of setting up another repo for this simple task though. We'd end up making a repo full of utility console commands that would have to drag in dependencies like `laminas-view` for a single command even if the project wasn't using it.

As a user, who installs laminas-view and also laminas-cli, this provides a 'batteries included' utility IMO that is more important than the pain of `symfony-console` as a dev dependency.

gsteel:
@Xerkus - there is also more discussion in #218

froschdesign:
> Imo this should be in a satellite package with explicit dependencies.

That is a spiral: we have more and more packages to manage, so I would approach this with caution. It might be possible to create a package for commands that contains everything across components but only for development.

I personally like the pragmatical solution like in https://github.com/laminas/laminas-view/issues/218#issuecomment-3252209445.

gsteel:
@Xerkus and @froschdesign I've added a conditional so that CLI config is only exposed for users that have installed symfony/cli.

Can we put this one to bed? It'd be nice to get v3 released 😁

froschdesign (approved):
I am fine with this! 👍🏻

PR #367, opened 2025-09-21
Remove requirements for unused PHP extensions

gsteel:
Updates `composer.json` to remove dependencies on extensions that are not used in `src`.

Also refreshes all locks

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,8 +34,6 @@
     },
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
-        "ext-dom": "*",
-        "ext-filter": "*",
         "ext-json": "*",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -34,8 +34,6 @@
     },
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
-        "ext-dom": "*",
-        "ext-filter": "*",
         "ext-json": "*",
```
Xerkus, line 37: should ext-json be kept since it is built in for a long time now and have to be explicitly disabled during compilation to be unavailable?

gsteel: IMO, leaving it in is fine, even though it's available by default, PHPStorm for example will moan about it not being present, and also composer require checker will fail too.

Ocramius (approved):

PR #371, opened 2025-09-24
Fixes order in navigation configuration for version 3 docs in `mkdocs.yml`

froschdesign:
Otherwise, the wrong version would be displayed.


gsteel (approved):
Thanks @froschdesign

PR #372, opened 2025-09-25
Refine types for view variables to `iterable<non-empty-string, mixed>`

gsteel:

PR #373, opened 2025-09-25
Allow for View Model Mutations Prior to Rendering

gsteel:
This affordance is primarily for Mezzio so that it becomes feasible to apply default parameters/variables to view models nested with arbitrary depth, without having to duplicate most of the code in `Laminas\View\View`.

From the context of Mezzio Laminas View Renderer, it would do something like:

```php
$this->renderer->registerPreRenderHandler(function (ModelInterface $model): ModelInterface {
    return $this->mergeViewModel($model->getTemplate(), $model);
});
```

This branch is cut from #372 so that will need merging first…

### DIFF
--- a/src/Renderer/PhpRenderer.php
+++ b/src/Renderer/PhpRenderer.php
@@ -5,13 +5,13 @@ declare(strict_types=1);
 namespace Laminas\View\Renderer;
 
 use Laminas\View\Exception\RenderingFailedException;
-use Laminas\View\Helper\ViewModel;
+use Laminas\View\Helper\ViewModel as ViewModelHelper;
 use Laminas\View\HelperPluginManagerInterface;
 use Laminas\View\Model\ModelInterface;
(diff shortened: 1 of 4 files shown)


froschdesign:
Is there any usage outside of Mezzio that requires some documentation to be added?

gsteel:
@froschdesign 

I was thinking that it could fit in a cookbook section, but it'll look a bit lonely on its own.

The intention is that it's a feature open to usage outside of Mezzio as part of the public api.

It doesn't really belong on the quick start, so I was hoping we might skip the docs for now until its worth making those improvements for users. I don't think this feature will be used much as it's not necessary for normal usage of the lib

froschdesign:
> I don't think this feature will be used much as it's not necessary for normal usage of the lib.

I am fine with this!

froschdesign (approved):

PR #377, opened 2025-10-13
Drop support for PHP 8.1 and upgrade PHPUnit to 11.x

gsteel:
- Refreshes all locks
- Removes PHP 8.4 compat conflicts in favour of installing minimum versions of transient dependencies in dev
- Mark all classes as readonly where possible


Ocramius (approved):
LGTM

PR #381, opened 2025-10-22
Remove `HelperInterface` requiring all plugins to be `callable`

gsteel:
`HelperInterface` is empty and is purely an affordance for legacy helpers that never implemented an `__invoke` method.

Removing the interface and all affordances means that users will be forced to drop the interface and add at a minimum `public function __invoke(): self;`.

This patch is not necessary and the change could be delayed until v4


src/HelperPluginManager.php
```diff
@@ -171,30 +170,29 @@
...
 
     /**
-     * @template InstanceParam of HelperInterface
+     * @template InstanceParam
      * @param class-string<InstanceParam>|string $id Service name of plugin to retrieve.
-     * @return ($id is class-string<InstanceParam> ? InstanceParam : InstanceType)
+     * @return ($id is class-string<InstanceParam> ? InstanceParam : callable)
      */
-    public function get(string $id): mixed
+    public function get(string $id): callable
```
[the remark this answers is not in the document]
Xerkus: Huh.

@gsteel did you enable it here or did it get enabled by default?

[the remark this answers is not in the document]
gsteel: Been exporting sarif results to GH in v3 for a while - I enabled it.

This is one of 6 issues already in the baseline.

src/HelperPluginManager.php
```diff
@@ -171,30 +169,29 @@ public function __construct(
...
             );
         }
     }
 
     /**
-     * @template InstanceParam of HelperInterface
+     * @template InstanceParam
      * @param class-string<InstanceParam>|string $id Service name of plugin to retrieve.
-     * @return ($id is class-string<InstanceParam> ? InstanceParam : InstanceType)
+     * @return ($id is class-string<InstanceParam> ? InstanceParam : callable)
```
Xerkus, line 185: Will this work?

```suggestion
     * @return ($id is class-string<InstanceParam> ? InstanceParam&callable : callable)
```

gsteel: No - but I can narrow the template to `object&callable` so it returns `object&callable|callable` - and psalm will correctly infer the object type too.

Xerkus: It does not matter. It needs to be on the interface to be useful

gsteel: Forgot the interface… done.

src/Renderer/Template.php
```diff
@@ -136,15 +135,10 @@ public function __call(string $method, array $args): mixed
             throw RenderingFailedException::becauseOfAnUnknownPlugin($method, $this->__template, $e);
         }
 
-        assert(is_callable($plugin) || $plugin instanceof HelperInterface);
+        assert(is_callable($plugin));
```
Xerkus, line 138: Change on the interface did not really do anything.  Can be checked here with psalm-trace: we still get an inferred `callable|mixed` hence the need for the assert.

psalm does not handle `@template T extends T1` that is on service manager plugin manager interface nor does it handle `@template T of T1`. It worked in HelperPluginManager because constraint uses actual types in constraint

Xerkus:
For the record.

Alternative path I proposed was to require all helpers to implement `HelperInterface` as a marker with no methods. At the same time `Template::__call()` would be modified to always return instance with no attempt at calling it. In fact, passing any parameters would cause a fail.

The migration for callable helpers that were doing anything else beside return self in their `__invoke()` would be 

```php
// old. Will throw.
echo $this->helperName($params);

// new
echo $this->helperName()($params);
// explicit
echo $this->helperName()->__invoke($params);
```

This creates error prone problem for when there are no parameters to pass:
```php
// old
echo $this->doctype();

// new, erroneous noop
echo $this->doctype();

// new, correct:
echo $this->doctype()();
```

Where helpers were simply returning self nothing changes:
```php
// old and new
$this->headTitle()->prepend($title);
```

I have a slight preference for this approach.

@gsteel argued that view helpers are methods on `$this` (`Template` in 3.x and was `PhpRenderer` in 2.x) in view scripts and therefore requiring helper to be callable is more sensible approach. Thus `HelperInterface` is redundant.

For existing non-callable `HelperInterface` helpers introduction of `public function __invoke(): self { return $this; }` while dropping the interface would be sufficient to keep support for both major versions. That is assuming there were no conflicts with other breaking changes.

src/Helper/HelperInterface.php
```diff
@@ -1,9 +0,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace Laminas\View\Helper;
-
-interface HelperInterface
-{
-}
```
gsteel, line 9: `HelperInterface` may still be useful for the purposes of DI.

It would be much easier to create a one-size-fits-all reflection-based abstract factory, if you can call `is_a($class, HelperInterface::class, true)` on a reflection parameter and therefore fetch it from the helper plugin manager.

gsteel:
Closing…

- I think helpers are working fine as they are
- Changing things even more is just more grief for users
- I don't have the appetite for more changes here. It's time this was released

PR #382, opened 2025-10-23
Add un-prefixed `@internal` where necessary

gsteel:
Adding `@internal` will signal intent to a wider variety of tools. Keeping `@psalm-internal` with the allowed namespaces prevents Psalm issues in our code base

### DIFF
--- a/src/Helper/Placeholder/Container.php
+++ b/src/Helper/Placeholder/Container.php
@@ -23,6 +23,8 @@ use function ob_start;
  *
  * This class is not part of the public API and has no BC guarantees
  *
+ * @internal
+ *
  * @template TValue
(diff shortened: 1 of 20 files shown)


src/Helper/Placeholder/Container.php
```diff
@@ -23,6 +23,8 @@
  *
  * This class is not part of the public API and has no BC guarantees
  *
+ * @internal
```
froschdesign, line 26: If I need the same functions, should I create this class again?

gsteel: This class specifically could be made public if we think users would benefit from it, but I'd rather keep it as an internal implementation detail until users say they want it - we can remove `@internal` easily, but adding it later is breaking BC

PR #383, opened 2025-10-23
Prevent ambiguous arguments to `PhpRenderer::render`

gsteel:
`PhpRenderer::render` should be called with `(ViewModel)` or `(string, iterable)`.

When called with `(ViewModel, iterable)`, an exception will now be thrown instead of silently ignoring the `iterable` representing the view variables.

This will be useful signalling to users because in previous versions, the view model and the given variables would have been merged.

### DIFF
--- a/src/Exception/RenderingFailedException.php
+++ b/src/Exception/RenderingFailedException.php
@@ -105,7 +105,7 @@ final class RenderingFailedException extends RuntimeException
     public static function becauseOfAmbiguousArgumentsToPhpRenderer(): self
     {
         return new self(
-            'Passing both view model view variables to render is ambiguous. '
+            'Passing both view model and view variables to render is ambiguous. '
             . 'Either provide just the model, or, a template name and the variables with '
             . 'which to create the model.'
         );
--- a/src/Exception/RenderingFailedException.php
+++ b/src/Exception/RenderingFailedException.php
@@ -101,4 +101,13 @@ final class RenderingFailedException extends RuntimeException
             $templatePath,
         ), 0, $previous);
     }
+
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
From my PoV, this is OK, but I would also explore restricting this at type level, somehow

src/Renderer/PhpRenderer.php
```diff
@@ -50,6 +52,13 @@ public function render(
             throw RenderingFailedException::becauseATemplateWasNotSpecified();
         }
 
+        $variablesEmpty = (is_object($variables) ? iterator_to_array($variables, false) : $variables) === []
+            || $variables === null;
+
+        if ($templateNameOrModel instanceof ModelInterface && ! $variablesEmpty) {
+            throw RenderingFailedException::becauseOfAmbiguousArgumentsToPhpRenderer();
+        }
+
```
Xerkus: I think it can be stricter and require null.
Something passed at all is indicative of a problem. It just happened to be empty at a time.

src/Exception/RenderingFailedException.php
```diff
@@ -101,4 +101,13 @@ public static function becauseOfAnUnknownPlugin(
             $templatePath,
         ), 0, $previous);
     }
+
+    public static function becauseOfAmbiguousArgumentsToPhpRenderer(): self
+    {
+        return new self(
+            'Passing both view model view variables to render is ambiguous. '
```
Xerkus: Typo?

```suggestion
            'Passing both view model and variables to render is ambiguous. '
```

Xerkus (approved):

PR #386, opened 2025-10-23
Remove the `RenderChildModel` View Helper

### LINKED ISSUE #385: Deprecate and Remove the `RenderChildModel` Helper
This helper only renders the first child with a given name, cannot render recursively and does not consider `append` flags.

It's a relic and should be removed.

gsteel:
Closes #385


docs/book/v3/migration/v2-to-v3.md
```diff
@@ -543,6 +543,12 @@ The deprecated navigation view helpers such as `Breadcrumbs`, and `Menu` etc hav
 
 As such, the namespace for these helpers has changed from `Laminas\View\Navigation` to `Laminas\Navigation\View\Helper`, so if you have referenced the FQCNs of these helpers in your code, you will need to update them accordingly.
 
+#### RenderChildModel
+
+The `RenderChildModel` helper never adequately rendered nested views, or took into account appending views.
+It was also un-documented, and not likely used very much.
+Possible use-cases were dubious at best and rendering view models can more easily be achieved by using the `Partial` view helper.
```
Xerkus: Partial helper is not the same functionality.
I would suggest something like this:

```suggestion
Use case was for manual rendering of nested view models with recursive tree
rendering disabled. Due to a bug in the initial implementation it was never
consistent with the recursive tree rendering.

Its functionality is entirely replaced with regular capture functionality of
nested view models which can no longer be disabled.
```

Fore reference, in the v2 docs nested view models and capture functionality is covered in Quick Start https://docs.laminas.dev/laminas-view/v2/quick-start/#nesting-view-models

Xerkus (approved):

PR #387, opened 2025-10-23
Deprecate the `RenderChildModel` Helper

gsteel:
Closes #386

### DIFF
--- a/src/Helper/RenderChildModel.php
+++ b/src/Helper/RenderChildModel.php
@@ -16,6 +16,9 @@ use function sprintf;
  * Finds children matching "capture-to" values, and renders them using the
  * composed view instance.
  *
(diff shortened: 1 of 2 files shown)


src/Helper/RenderChildModel.php
```diff
@@ -16,6 +16,9 @@
  * Finds children matching "capture-to" values, and renders them using the
  * composed view instance.
  *
+ * @deprecated Since 2.44.0. This helper will be removed in 3.0.0. It's functionality should be replicable by using
+ *             the Partial view helper
```
Xerkus: It is not really replicable by partial but rather by regular capture-to with tree rendering that will always be used in 3.0

Xerkus (approved):

PR #388, opened 2025-10-23
Add defensive check for empty strings to `ViewModel::setCaptureTo()`

### LINKED ISSUE #384: Ensure `ViewModel::captureTo()` cannot be an empty string at runtime
Either check relevant tests exist to ensure using an empty string will blow up, or, add runtime checks to the setter.

Only the setter is capable of making this an empty string

gsteel:
Closes #384

### DIFF
--- a/src/Model/ViewModel.php
+++ b/src/Model/ViewModel.php
@@ -217,7 +217,7 @@ final class ViewModel implements ModelInterface, ClearableModelInterface, Retrie
     {
         /** @psalm-suppress TypeDoesNotContainType Adding a defensive check here regardless of documented types */
         if ($capture === '') {
-            throw new InvalidArgumentException('The `capture` name cannot be an empty string');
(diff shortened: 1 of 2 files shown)


src/Model/ViewModel.php
```diff
@@ -214,6 +215,11 @@ public function getChildrenByCaptureTo(string $capture, bool $recursive = true):
 
     public function setCaptureTo(string $capture): static
     {
+        /** @psalm-suppress TypeDoesNotContainType Adding a defensive check here regardless of documented types */
+        if ($capture === '') {
+            throw new InvalidArgumentException('The `capture` name cannot be an empty string');
```
Xerkus: Minor nitpick 
```suggestion
            throw new InvalidArgumentException('The `capture` target cannot be an empty string');
```

Xerkus (approved):

PR #390, opened 2025-10-23
Introduce View Interface

gsteel:


gsteel:
Closing - We still need to introduce a view interface, but all the other changes here can get in the bin.

PR #411, opened 2026-06-01
Fix (docs) argument order in view render method

pine3ree:
Quick start documentation error.

Reference:

https://github.com/laminas/laminas-view/blob/c7f8af6e20a46306fe577cf730474a337ce4d1fd/src/View.php#L48-L52


gsteel:
Thanks very much @pine3ree 👍

PR #416, opened 2026-06-02
Add support for `laminas-translator@2`

gsteel:
Adds a factory for tests to produce a v1 or v2 stub by inspecting the method signature of the installed interface. It's kinda messy, but means we can continue to test using both versions.


Ocramius (approved):
LGTM 👍

Ocramius:
Thanks @gsteel!

